PR #132, opened 2021-07-21
fix: changes per audit tickets ERE01, 02 and 03. Stlye changes accord…

robrobbins:
…ingly


JTraversa (commented):
Perc20 -> PErc20 matches a bit better with the rest of our codebase (e.g. IErc20, IZcToken)

Outside that, everything looks good to me

PR #167, opened 2021-09-20
implement storing by hash vs key for initates, exits and cancels

robrobbins:


JTraversa (approved):
Looks right

JTraversa (commented):
Edit: reposted the same thing accidentally woops

PR #169, opened 2021-10-24
Gas optimizations

JTraversa:
~25 commits across gas optimizations 

One noteworthy change:
Mass approval for underlying -> compound tokens:
We will need to mass approve all underlying -> cToken addresses on deployment. This saves marginal approvals on each call.

Others include math changes, better caching, redundant requires, etc, folding maturityrate into the market and removing maturity bool.

### DIFF
--- a/test/marketplace/MarketPlace.sol
+++ b/test/marketplace/MarketPlace.sol
@@ -133,7 +133,7 @@ contract MarketPlace {
     
     if (mkt.maturityRate == 0) {
       // Attempt to Mature it
-      require(matureMarket(u, m), 'failed to mature the market');
+      matureMarket(u, m);
     }
 
     // Burn user's zcTokens
--- a/test/marketplace/MarketPlace.sol
+++ b/test/marketplace/MarketPlace.sol
@@ -168,12 +168,8 @@ contract MarketPlace {
   /// @param m Maturity timestamp of the market
   /// @param a Amount of zcTokens being redeemed
   function calculateReturn(address u, uint256 m, uint256 a) internal returns (uint256) {
-    // Calculate difference between the cToken exchange rate @ maturity and the current cToken exchange rate
-    uint256 yield = ((CErc20(markets[u][m].cTokenAddr).exchangeRateCurrent() * 1e26) / maturityRate[u][m]) - 1e26;
-    uint256 interest = (yield * a) / 1e26;
-
-    // Calculate the total amount of underlying principle to return
-    return a + interest;
+    uint256 rate = CErc20(markets[u][m].cTokenAddr).exchangeRateCurrent();
+    return  a * rate / maturityRate[u][m];
   }
 
   function cTokenAddress(address a, uint256 m) external view returns (address) {
--- a/test/marketplace/MarketPlace.sol
+++ b/test/marketplace/MarketPlace.sol
@@ -15,10 +15,10 @@ contract MarketPlace {
     address cTokenAddr;
     address zcTokenAddr;
     address vaultAddr;
+    uint256 maturityRate;
   }
 
   mapping (address => mapping (uint256 => Market)) public markets;
-  mapping (address => mapping (uint256 => uint256)) public maturityRate;
 
   address public admin;
   address public swivel;
@@ -68,7 +68,7 @@ contract MarketPlace {
     // TODO can we live with the factory pattern here both bytecode size wise and CREATE opcode cost wise?
     address zctAddr = address(new ZcToken(u, m, n, s, d));
     address vAddr = address(new VaultTracker(m, c, swivel));
-    markets[u][m] = Market(c, zctAddr, vAddr);
+    markets[u][m] = Market(c, zctAddr, vAddr, 0);
 
     emit Create(u, m, c, zctAddr, vAddr);
 
@@ -79,12 +79,15 @@ contract MarketPlace {
   /// @param u Underlying token address associated with the market
   /// @param m Maturity timestamp of the market
   function matureMarket(address u, uint256 m) public returns (bool) {
-    require(maturityRate[u][m] > 0, 'market already matured');
+    
+    Market memory mkt = markets[u][m];
+    
+    require(mkt.maturityRate > 0, 'market already matured');
     require(block.timestamp >= ZcToken(markets[u][m].zcTokenAddr).maturity(), "maturity not reached");
 
     // Set the base maturity cToken exchange rate at maturity to the current cToken exchange rate
-    uint256 currentExchangeRate = CErc20(markets[u][m].cTokenAddr).exchangeRateCurrent();
-    maturityRate[u][m] = currentExchangeRate;
+    uint256 currentExchangeRate = CErc20(mkt.cTokenAddr).exchangeRateCurrent();
+    markets[u][m].maturityRate = currentExchangeRate;
 
     // Set Floating Market "matured" to true
     require(VaultTracker(markets[u][m].vaultAddr).matureVault(currentExchangeRate), 'maturity not reached');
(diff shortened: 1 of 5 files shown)


test/marketplace/MarketPlace.sol
```diff
@@ -15,13 +15,12 @@ contract MarketPlace {
...
     address zcTokenAddr;
     address vaultAddr;
+    uint256 maturityRate;
   }
 
   mapping (address => mapping (uint256 => Market)) public markets;
-  mapping (address => mapping (uint256 => bool)) public mature;
-  mapping (address => mapping (uint256 => uint256)) public maturityRate;
 
-  address public immutable admin;
```
robrobbins, line 24: why? so that it can be changed? wut?

test/marketplace/MarketPlace.sol
```diff
@@ -39,14 +38,14 @@ contract MarketPlace {
   }
 
   /// @notice Allows the admin to transfer ownership
-  /// @param t
-  function transferAdmin(address t) external onlyAdmin(admin) returns (bool) {
+  /// @param t Address admin is being transferred to 
+  function transferAdmin(address t) external onlyAddress(admin) returns (bool) {
```
robrobbins, line 42: oh, so that it can be delegated to another address? we aint gunna use `onlyAddress` tho..

why not just leave it? what's wrong with calling the authorized address admin? ... ic... you want to normalize it across contracts, ok...

just use `authorized` then. we can drop the `only`

test/marketplace/MarketPlace.sol
```diff
@@ -64,12 +63,13 @@ contract MarketPlace {
...
-    require(swivel != address(0), 'swivel contract address not set');
+  ) public onlyAddress(admin) returns (bool) {
+    address swivelAddr = swivel;
+    require(swivelAddr != address(0), 'swivel contract address not set');
     // TODO can we live with the factory pattern here both bytecode size wise and CREATE opcode cost wise?
     address zctAddr = address(new ZcToken(u, m, n, s, d));
-    address vAddr = address(new VaultTracker(m, c, swivel));
-    markets[u][m] = Market(c, zctAddr, vAddr);
+    address vAddr = address(new VaultTracker(m, c, swivelAddr));
+    markets[u][m] = Market(c, zctAddr, vAddr, 0);
```
robrobbins, line 72: there's a 4th arg?

test/marketplace/MarketPlace.sol
```diff
@@ -80,17 +80,17 @@ contract MarketPlace {
...
+    // set the base maturity cToken exchange rate at maturity to the current cToken exchange rate
+    uint256 currentExchangeRate = CErc20(mkt.cTokenAddr).exchangeRateCurrent();
+    markets[u][m].maturityRate = currentExchangeRate;
 
-    // Set Floating Market "matured" to true
-    require(VaultTracker(markets[u][m].vaultAddr).matureVault(), 'maturity not reached');
+    // set Floating Market "matured" to true
+    VaultTracker(mkt.vaultAddr).matureVault(currentExchangeRate);
 
     emit Mature(u, m, block.timestamp, currentExchangeRate);
```
robrobbins, line 95: arg order? ;)

test/marketplace/MarketPlace.sol
```diff
@@ -64,12 +63,13 @@ contract MarketPlace {
     string memory n,
     string memory s,
     uint8 d
-  ) public onlyAdmin(admin) returns (bool) {
-    require(swivel != address(0), 'swivel contract address not set');
+  ) public onlyAddress(admin) returns (bool) {
```
robrobbins, line 66: `public authorized(admin)`... we can normalize the contracts to use that

test/swivel/Sig.sol
```diff
@@ -37,7 +37,7 @@ library Sig {
   /// @return v The verification bit
   /// @return r First 32 bytes
   /// @return s Next 32 bytes
-  function split(bytes memory sig) internal pure returns (uint8, bytes32, bytes32) {
+  function split(bytes memory sig) private pure returns (uint8, bytes32, bytes32) {
```
robrobbins, line 40: we have many `internal` methods. if you are worried about derived contracts then why not change them all to `private`. seems half-baked to just change 1?

test/swivel/Swivel.sol
```diff
@@ -195,24 +191,26 @@ contract Swivel {
   /// @param a Array of order volume (principal) amounts relative to passed orders
   /// @param c Components of a valid ECDSA signature
   function exit(Hash.Order[] calldata o, uint256[] calldata a, Sig.Components[] calldata c) external returns (bool) {
-    for (uint256 i=0; i < o.length; i++) {
-      // Determine whether the order being filled is an exit
-      if (!o[i].exit) {
-        // Determine whether the order being filled is a vault initiate or a zcToken initiate
-          if (!o[i].vault) {
-            // If filling a zcToken initiate with an exit, one is exiting zcTokens
+    unit256 length = o.length;
```
robrobbins, line 194: its already calldata man. is this really doing anything? 

besides just being ugly

i kind of see where avoiding storage can save _some_ gas, but not avoiding _calldata_. i'd need proof here. calling bs... maybe i learn something tho

test/swivel/Swivel.sol
```diff
@@ -396,6 +394,17 @@ contract Swivel {
     return true;
   }
 
+  /// @notice Allows the admin to mass approve all compound markets, saving marginal approvals
+  /// @param u The underlying token address
+  /// @param c The compound cToken address
+  function approveCompound(address[] calldata u, address[] calldata c) external onlyAdmin(admin) returns (bool) {
+    for (uint256 i; i < u.length; i++) {
```
robrobbins, line 401: see, this one is calldata u, not hoisted...

test/vaulttracker/VaultTracker.sol
```diff
@@ -128,16 +135,14 @@ contract VaultTracker {
...
 
-  /// @notice ...
-  function matureVault() external onlyAdmin(admin) returns (bool) {
-    require(!matured, 'already matured');
-    require(block.timestamp >= maturity, 'maturity has not been reached');
-    matured = true;
-    maturityRate = CErc20(cTokenAddr).exchangeRateCurrent();
+  /// @notice Matures the vault
+  /// @param currentExchangeRate the current cToken exchangeRate
+  function matureVault(uint256 currentExchangeRate) external onlyAdmin(admin) returns (bool) {
```
robrobbins, line 140: arg should be single letter

robrobbins (changes requested):

test/marketplace/MarketPlace.sol
JTraversa: The ticket essentially pointed out that we have both onlyAdmin and onlySwivel, and that we could just have one that uses the address as a param, instead of seperate modifiers (saving bytecode)

test/marketplace/MarketPlace.sol
JTraversa: Yea, one change involved removing the entire mapping for both `matured` and `maturityRate`, folding `maturityRate` into the market.

This lets us check if the market has matured by comparing `maturityRate` to the base case of 0, and also saves gas from using 1 mapping instead of 3.

test/marketplace/MarketPlace.sol
JTraversa: 👍

test/swivel/Swivel.sol
JTraversa: i actually commented *exactly* this on the audit and had to prove it doing my own remix tests.

JTraversa: Woops wasn't this exact change but i tested a few and all were actually minor improvements. 

(this was the one i was thinking of: https://github.com/code-423n4/2021-09-swivel-findings/issues/43)

test/swivel/Swivel.sol
JTraversa: Ahhh thats cause this wasnt audited hah, this is my own code for the new approval workflow on deployment (avoiding approvals on every trade)

JTraversa: Can/should add the same optimization to it even though were the only ones calling the function

test/marketplace/MarketPlace.sol
robrobbins: done

PR #170, opened 2021-10-24
Med risk

JTraversa:
- prevent overwriting already existing markets
- fix double fee take in zcTokenexitzcTokeninit


robrobbins (commented):
looks ok to me i think

robrobbins:
address in #181

PR #174, opened 2021-11-03
Low risk

JTraversa:


test/marketplace/MarketPlace.sol
```diff
@@ -17,11 +17,12 @@ contract MarketPlace {
     address vaultAddr;
     uint256 maturityRate;
   }
-
+  /// @dev maps a token to a given maturity and from there to a given market in our marketplace
   mapping (address => mapping (uint256 => Market)) public markets;
 
   address public admin;
   address public swivel;
+  bool private swivelInitiated;
```
robrobbins, line 25: we don't need this bool.

test/marketplace/MarketPlace.sol
```diff
@@ -46,24 +47,26 @@ contract MarketPlace {
   
   /// @param s Address of the deployed swivel contract
   function setSwivelAddress(address s) external onlyAddress(admin) returns (bool) {
+    require (!swivelInitiated, 'can only set the swivel address once`);
```
robrobbins, line 50: we can require that the adress 'swivel' == `0x0` in order to be set (meaning: it hasn't been set yet). once it has this check will fail. use the .sol default instead of an extra bool

test/marketplace/MarketPlace.sol
```diff
@@ -46,24 +47,26 @@ contract MarketPlace {
...
     uint256 m,
     address c,
     string memory n,
-    string memory s,
-    uint8 d
+    string memory s
   ) public onlyAddress(admin) returns (bool) {
     address swivelAddr = swivel;
     address uAddr = CErc20(c).underlying(); 
+    uint8 d = Erc20(uAddr).decimals();
```
robrobbins, line 69: this is gunna echo out everywhere, weeeee

test/marketplace/MarketPlace.sol
```diff
@@ -92,7 +95,7 @@ contract MarketPlace {
     // set Floating Market "matured" to true
     VaultTracker(mkt.vaultAddr).matureVault(currentExchangeRate);
 
-    emit Mature(u, m, block.timestamp, currentExchangeRate);
+    emit Mature(u, m, currentExchangeRate, block.timestamp);
```
robrobbins, line 98: how did we miss that? prob added it late or something

test/marketplace/Abstracts.sol
```diff
@@ -2,17 +2,18 @@
 
 pragma solidity 0.8.4;
 
-abstract contract Erc20 {
-	function approve(address, uint256) virtual external returns (bool);
-	function transfer(address, uint256) virtual external returns (bool);
-	function balanceOf(address) virtual external returns (uint256);
-	function transferFrom(address, address, uint256) virtual public returns (bool);
+interface Erc20 {
```
robrobbins, line 5: is this change to an interface because of the auto-generated getter for decimals? does it fail if you just put decimals in the abstract?

test/marketplace/Abstracts.sol
```diff
@@ -2,17 +2,18 @@
...
+	function approve(address, uint256) external returns (bool);
+	function transfer(address, uint256) external returns (bool);
+	function balanceOf(address) external view returns (uint256);
+	function transferFrom(address, address, uint256) external returns (bool);
+	function decimals() external returns (uint8);
```
robrobbins, line 10: if we keep interface then this should be view as well

robrobbins (changes requested):

test/marketplace/MarketPlace.sol
JTraversa: Yea thats the way better solution. Good call.

test/marketplace/Abstracts.sol
JTraversa: Just gonna reference the ticket: https://github.com/code-423n4/2021-09-swivel-findings/issues/93

test/marketplace/Abstracts.sol
JTraversa: 👍

test/marketplace/Abstracts.sol
robrobbins: could be correct, lemme take a closer look. idr why i made them abstracts to begin with. if there's no reason then sure, interface prob the correct call as long as it compiles and runs correctly

test/marketplace/MarketPlace.sol
robrobbins: done as part of #179

test/marketplace/MarketPlace.sol
robrobbins: done. #179

test/marketplace/MarketPlace.sol
robrobbins: done. #178

test/marketplace/Abstracts.sol
robrobbins: done #178

test/marketplace/Abstracts.sol
robrobbins: #178

robrobbins:
rest is addressed in #181

PR #180, opened 2021-11-15
paused circuit breaker implemented. double fee take removed. no trans…

robrobbins:
…fer notional to self (high risk tickets)


JTraversa (commented):
With a pause method, think we should have an unpause method?d

In most/any use cases where you pause, you likely dont want to unpause?

lmk if you think of any

test/swivel/Swivel.sol
```diff
@@ -272,17 +272,17 @@ contract Swivel {
...
-    uint256 fee = ((premiumFilled * 1e18) / fenominator[3]) / 1e18;
+    uint256 premiumFilled = (a * o.premium) / o.principal;
 
     Erc20 uToken = Erc20(o.underlying);
-    // transfer premium minus fee from maker to sender
-    uToken.transferFrom(o.maker, msg.sender, premiumFilled - fee);
+    // transfer premium from maker to sender
+    uToken.transferFrom(o.maker, msg.sender, premiumFilled);
 
+    uint256 fee = ((premiumFilled * 1e18) / fenominator[3]) / 1e18;
```
JTraversa, line 282: I'll move fee calc up to 276 just for code readability/consistency sake at the end of the audit review process unless any reason otherwise. I just like having the math one after another in each Swivel method.

JTraversa (approved):

test/swivel/Swivel.sol
robrobbins: looks back at other methods..

yeah, its dumb everywhere now. we really shouldn't spend the resources calculating and storing a variable that isn't immediately used and may not be used ah all (if the `transferFrom` after it fails, this var was never needed to begin with)

JTraversa (approved):
🚢

PR #181, opened 2021-11-16
Fix/med risk

robrobbins:
note: event names were never past tensed in our contracts. Most are the method name "upcased", some are shorter if possible.


robrobbins:
also normalized feenominators / feenominator

JTraversa (approved):
👍 

Nothing wrong w/ any of this. I just dont see anything for this med risk issue: https://github.com/code-423n4/2021-09-swivel-findings/issues/97

Though that could obviously be a different PR

PR #205, opened 2022-04-29
test: unit test for mint method of zc token

sourabhmarathe:
This PR creates a basic unit test of the zc token interface by using the mocked ZC token.

### DIFF
--- a/pkg/lendertesting/zc_token_test.go
+++ b/pkg/lendertesting/zc_token_test.go
@@ -1,0 +1,65 @@
+package lendertesting
+
+import (
+	"math/big"
+	test "testing"
+
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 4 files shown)


pkg/lendertesting/zc_token_test.go
```diff
@@ -0,0 +1,65 @@
+package lendertesting
```
robrobbins: just like YieldToken -- this should be in `/testing` as its testing a /mock/ specifically

robrobbins (changes requested):

pkg/testing/yield_token_test.go
```diff
@@ -1,4 +1,4 @@
-package lendertesting
+package testing
```
robrobbins, line 1: wut? was yield_token_test in the `lendertesting`? well it _should_ be in `testing` so all good...

robrobbins (approved):

PR #216, opened 2022-05-05
Implement lend for pendle

sourabhmarathe:
This PR implements the lend method based on the contents of the `split-contracts` branch in the illuminate repo. It creates two interfaces: `IPendle` and `ISushi`, mocks them, and unit tests the a new `lend` override specific to the function signature of the Pendle protocol.


sourabhmarathe:
We are using this method from Pendle to determine the market's underlying & maturity.

https://github.com/pendle-finance/pendle-core/blob/master/contracts/interfaces/IPendleForge.sol#L199

PR #218, opened 2022-05-05
remove marketplace. implement those methods in illuminate. establish …

robrobbins:
…illuminatetesting domain


test/illuminate/Illuminate.sol
```diff
@@ -2,48 +2,95 @@
...
+    require(markets[u][m][uint256(Principals.Illuminate)] == address(0), 'market exists'); 
+
+    // deploy an illuminate token with this new market NOTE: ATM is using name as symbol args
+    address iToken = address(new ZcToken(u, m, n, s, d));
+
+    // the market will have the illuminate principal as its zeroth item, thus t should have Principals[1] as [0]
+    // TODO we could choose to put illuminate last in
+    address[8] memory market = [iToken, t[0], t[1], t[2], t[3], t[4], t[5], t[6]];
+
+    uint256 max = 2**256 - 1;
```
sourabhmarathe, line 58: We should store this value as a constant in `Constants.sol`. 
e.g. `uint256 MAX_UINT256 = uint256(-1)`
This will be slightly more efficient than computing the maximum value via exponentiation.

test/illuminate/Illuminate.sol
```diff
@@ -2,48 +2,95 @@
...
-    ZcToken(market[mPlace.Principals.Illuminate]).mint(msg.sender, a);
+    // ZcToken(market[mPlace.Principals.Illuminate]).mint(msg.sender, a);
 
-    emit Mint(p, u, m, a);
+    // emit Mint(p, u, m, a);
 
-    return true;
+    // return true;
+  // }
+
```
sourabhmarathe, line 91: Remove commented out code.

test/illuminate/Interfaces.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: UNLICENSED
+
+pragma solidity 0.8.13;
+
+interface IErc20 {
+	function approve(address, uint256) external returns (bool);
+	// function transfer(address, uint256) external returns (bool);
+	// function balanceOf(address) external returns (uint256);
+	// function transferFrom(address, address, uint256) external returns (bool);
```
sourabhmarathe, line 9: We'll need these methods for `lend` at least - why are they commented out here?

test/illuminate/Safe.sol
```diff
@@ -0,0 +1,123 @@
...
+
+      // Call the token and store if it succeeded or not.
+      // We use 68 because the calldata length is 4 + 32 * 2.
+      result := call(gas(), e, 0, pointer, 68, 0, 0)
+    }
+
+    require(success(result), "approve failed");
+  }
+
+  /// @param e Erc20 token to execute the call with
```
sourabhmarathe, line 37: Several methods commented out - was this intentional?

test/illuminate/ZcToken.sol
```diff
@@ -0,0 +1,98 @@
...
+    // address to;
+    // uint256 amount;
+  // }
+
+  // mapping of arguments sent to burn. key is the passed in address.
+  // mapping (address => uint256) public burnCalled;
+  // mapping of arguments sent to mint. key is the passed in address.
+  // mapping (address => uint256) public mintCalled;
+  // mapping of arguments sent to transferFrom. key is passed from address.
+  // mapping (address => TransferFromArgs) public transferFromCalled;
```
sourabhmarathe, line 22: Remove comments.

test/illuminate/ZcToken.sol
```diff
@@ -0,0 +1,98 @@
...
+    // we can set the privates in the constructor as well...
+    underlyingReturn = u;
+    maturityReturn = m;
+
+    name = n;
+    symbol = s;
+    decimals = d;
+  }
+
+  // function burnReturns(bool b) public {
```
sourabhmarathe, line 51: Unused methods/commented out?

sourabhmarathe (changes requested):
Mostly has to with removing stale code, otherwise LGTM

test/illuminate/Safe.sol
robrobbins: yes. it is unknown exactly what methods this will need. uncomment them as needed, then when the contract is done - delete the ones never used.

test/illuminate/ZcToken.sol
robrobbins: no. not yet. see above.

test/illuminate/Illuminate.sol
robrobbins: making a constants file not a pattern we've used before -- typically we'd just make an actual constant in the contract if it's used a bunch. don't think this is? i can take a look tho

test/illuminate/Illuminate.sol
robrobbins: LOL. yes. but. you are actually going to do it. why? this `mint` method is going to be moved into the `lender`

just copy it from here.

PR #226, opened 2022-05-10
Rename interfaces to match protocol and token pattern

sourabhmarathe:


robrobbins (dismissed):
these look right.

test/lender/Lender.sol
```diff
@@ -9,27 +9,27 @@ import './Safe.sol';
...
+  address public pendleAddr;
+  address public tempusAddr;
   address public senseToken;
 
   event Lend(uint8 principal, address indexed underlying, uint256 indexed maturity, uint256 returned);
   event Mint(uint8 principal, address indexed underlying, uint256 indexed maturity, uint256 amount);
 
   /// @param i the deployed Illuminate contract
-  constructor(address i, address s, address su, address t, address sa) {
+  constructor(address i, address s, address pr, address t, address st) {
```
robrobbins: so when we can't use the single letter paradigm, we use a full word. it's a one or the other kinda thing.

so probs `(i, swivel, p, t, sense)` -- we'll usually use full words for whatever would have clashed with one letter

test/lender/Lender.sol
```diff
@@ -9,27 +9,27 @@ import './Safe.sol';
...
     admin = msg.sender;
     illuminate = i; // TODO add an authorized setter for this?
     swivelAddr = s;
-    pendleRouter = su;
-    tempusRouter = t;
-    senseToken = sa;
+    pendleAddr = pr;
+    tempusAddr = t;
+    senseToken = st;
```
robrobbins: this should be `senseAddr`. the `...Token` naming schemes are used for `IFooToken` interfaces

test/lender/Lender.sol
```diff
@@ -187,11 +187,11 @@ contract Lender {
   function lend(uint8 p, address u, uint256 m, uint256 a, uint256 mb, uint256 d) public returns (uint256) {
       // Instantiate market and tokens
       address market = IIlluminate(illuminate).markets(u, m)[p];
-      IPendleToken pendle = IPendleToken(market);
+      IPendleToken pendleToken = IPendleToken(market); // rename to pendletoken
```
robrobbins: the var can just be token? `IPendleToken token = IPendleToken(principal)`

test/lender/Lender.sol
```diff
@@ -187,11 +187,11 @@ contract Lender {
   function lend(uint8 p, address u, uint256 m, uint256 a, uint256 mb, uint256 d) public returns (uint256) {
       // Instantiate market and tokens
       address market = IIlluminate(illuminate).markets(u, m)[p];
```
robrobbins: don't use `market` here this way -- a market is an array of addresses to Illuminate, this is a specific principal or principal token... so the var name could be like `pToken` or `principal`... we use token quite a bit (i think we used `eToken` for this in the element method).. i'd say do whatever we are doing elsewhere

we can prob come back and standardize this particular lookup to always be `principal` -- that's what it is.

test/lender/Lender.sol
```diff
@@ -176,7 +176,7 @@ contract Lender {
     return IElement(e).swap(swap, fund, r, d);
   }
 
-  /// @dev lend method signature for pendle
+  /// @dev lend method signature for pendleAddr
```
robrobbins: nah, thats being weird. this is for `Pendle` on a high level. Pendle as a product - as a protocol. focusing on the *Addr implementation detail is being over-specific

test/lender/Lender.sol
```diff
@@ -225,17 +226,17 @@ contract Lender {
   /// @param d deadline ?
   function lend(uint8 p, address u, uint256 m, uint256 a, uint256 r, address x, address t, uint256 d) public returns (uint256) {
       // Instantiate market and tokens
-      address market = IIlluminate(illuminate).markets(u, m)[p];
-      require(ITempus(market).yieldBearingToken() == IErc20Metadata(u), 'tempus underlying != underlying');
-      require(ITempus(market).maturityTime() == m, 'tempus maturity != maturity');
+      address tempusToken = IIlluminate(illuminate).markets(u, m)[p];
```
robrobbins: the whole `address principal` or `address tToken` thing again...

`tToken` follows the pattern that `Element...` set. again, i think we change them all to `principal` at some point...

robrobbins (changes requested):

test/lender/Lender.sol
sourabhmarathe: In this case, `senseToken` is used the same way other tokens are used. However, I think we are not using `senseToken` correctly. I am going to remove this reference and create a ticket to double check this.

test/lender/Lender.sol
sourabhmarathe: Agree, `principal` is a much better word for this.

robrobbins (approved):

PR #229, opened 2022-05-11
Feat/even moar market

robrobbins:


sourabhmarathe (approved):
Changes look good, I assume we'll be doing something similar for lender?

PR #230, opened 2022-05-11
Implement redeem method for APWine

sourabhmarathe:
This PR implements the `redeem` method for APWine. This redeem method shares a signature with `Tempus` and `Illuminate`, and as such, this PR also includes the beginning of the implementations of `redeem` for those methods (e.g. `TempusToken` work).


pkg/redeemertesting/redeem_test.go
```diff
@@ -88,6 +110,66 @@ func (s *redeemTestSuite) SetupSuite() {
...
+}
+
+func (s *redeemTestSuite) TestTempusRedeem() {
+
+}
+
+func (s *redeemTestSuite) TestIlluminateRedeem() {
+
+}
+
```
sourabhmarathe, line 168: Will be implemented in upcoming PRs.

pkg/redeemertesting/dep.go
```diff
@@ -72,17 +113,25 @@ func Deploy(e *Env) (*Dep, error) {
...
+		Illuminate:         mpContract,
+		RedeemerAddress:    redeemerAddress,
+		APWine:             apContract,
+		APWineAddress:      apAddress,
+		APWineToken:        aptContract,
+		APWineTokenAddress: aptAddress,
+		// Tempus:             tContract,
+		// TempusAddress:      tAddress,
+		// TempusToken:        ttContract,
+		// TempusTokenAddress: ttAddress,
```
sourabhmarathe, line 134: Will be uncommented in next Tempus redeem PR.

test/mocks/Tempus.sol
```diff
@@ -59,4 +62,11 @@ contract Tempus {
 
         return depositAndFixReturn;
     }
+
+    function redeemToBacking(address o, uint256 m, uint256 a, address u) external {
```
robrobbins, line 70: in methods that take more than 2 args we use a mapping whose first argument is the key whose value is map to hold the other args

https://github.com/Swivel-Finance/gost/blob/v2/test/mocks/Erc20.sol#L12=

test/mocks/Tempus.sol
```diff
@@ -28,6 +28,9 @@ contract Tempus {
     bool public isBackingTokenCalled;
     uint256 public minimumReturnCalled;
     uint256 public deadlineCalled;
+    address public ownerCalled;
```
robrobbins: use the pattern established in the mock Erc20 for how TransferFrom does this (vs using individual args saved separately) see below...

test/mocks/APWine.sol
```diff
@@ -11,6 +11,7 @@ contract APWine {
     uint256 public tokenOutCalled;
     uint256 public minimumAmountCalled;
     address public toCalled;
+    address public ownerCalled;
```
robrobbins: no no. these are not in pattern. look at the mock Erc20. 

1. `fooCalled` is always the name of a method that was called - not the name of an argument.
2. depending on the number of args to the method the stored value can be:

    * one arg -> the value is the arg
    * two args -> a mapping whose key is the first arg and the val is the 2nd
    * 2+ args -> a mapping whose key is arg 1 and the mapping holds args 2 -> N by name

test/mocks/ZcToken.sol
```diff
@@ -7,6 +7,8 @@ contract ZcToken {
     bool private mintReturn;
 
     mapping (address => uint256) public mintCalled;
+    address public ownerCalled;
```
robrobbins: samesies

robrobbins (dismissed):
I'd say merge in your changes.

then come back in a follow up to audit the mock style / patterns

robrobbins (approved):
now fixup them mocks yo...

PR #232, opened 2022-05-11
Audit all mocks to fit pattern

sourabhmarathe:


test/mocks/Element.sol
```diff
@@ -23,33 +23,24 @@ struct FundManagement {
...
-    address fundManagementSenderCalled;
-    uint256 singleSwapAmountCalled;
-
-    function deadline() external view returns (uint256) {
-        return deadlineCalled;
+contract Element {
+    struct SwapArgs {
+        address recipient;
+        uint256 swapAmount;
+        uint256 _return;
```
sourabhmarathe: Look into this

sourabhmarathe: Look into this

pkg/lendertesting/lend_test.go
```diff
@@ -372,7 +372,7 @@ func (s *lendTestSuite) TestLendElement() {
 	maturity := big.NewInt(100000)
 	elementPoolId := [32]byte{1}
 	amount := big.NewInt(10000)
-	returnAmount := big.NewInt(100)
```
robrobbins, line 375: returnAmount is ok -- that wasn't part of the `return_` joke from Element (or whomever's) garbage code

robrobbins: i mean that came from jule's `amount returned` comments... no idea what it actually should be called. maybe it _is_ actually `limit` -- think we are using `r` in the args...

robrobbins: you can always change those as you learn more

robrobbins (approved):

PR #233, opened 2022-05-12
Implement redeem function for tempus

sourabhmarathe:
This PR implements the redeem function for Tempus.


pkg/redeemertesting/redeem_test.go
```diff
@@ -159,7 +179,49 @@ func (s *redeemTestSuite) TestAPWineRedeem() {
 }
 
 func (s *redeemTestSuite) TestTempusRedeem() {
+	assert := assert.New(s.T())
+
+	amount := big.NewInt(1000)
+	maturity := big.NewInt(9999999)
+	adapter := common.HexToAddress("0x0000000000000000000000000000000000000002")
```
robrobbins: we should be cautious with how we refer to this arg.

its `o` in the contract method and represented as `owner...` . I realize that the 3 different principals that take that signature may use that address for different things - but we likely want to refer to it uniformly. 

I am loathe to allow the words `adapter` and `router` etc in to the vernacular (even in tests) unless we must. and then would should be able to limit it to one term.

robrobbins (dismissed):
looks good. 

just the comment about `adapter`

robrobbins (approved):

PR #237, opened 2022-05-13
Implement redeem for sense

sourabhmarathe:
Also includes the `Router` -> `Addr` change in `Redeemer.sol`.


test/redeemer/Redeemer.sol
```diff
@@ -15,19 +15,21 @@ contract Redeemer {
 
   address public admin;
   address public illuminate;
-  address public apwineRouter;
-  address public tempusRouter;
-  address public pendleRouter;
+
+  // TODO: Rename Router to Addr
```
robrobbins, line 19: i'd imagine you can remove this comment now?

robrobbins (approved):

PR #242, opened 2022-05-17
Review swivel lend and redeem implementation

sourabhmarathe:
There are several fixes and TODOs addressed for Swivel's implementation of `redeem` and `lend` in here. In addition, any question spots have been noted in the google doc created by Kevin.


test/lender/Lender.sol
```diff
@@ -102,31 +102,34 @@ contract Lender {
...
       Swivel.Order memory order = o[i];
       // Require the Swivel order provided matches the underlying and maturity market provided    
       require(order.maturity == m, 'swivel maturity != maturity');
       require(order.underlying == u, 'swivel underlying != underlying');
-      // Sum the total amount lent to Swivel (amount of zcb to mint)
+      // Sum the total amount lent to Swivel (amount of zc tokens to mint)
       lent += a[i];
       // Sum the total amount of premium paid from Swivel (amount of underlying to lend to yield)
-      returned += a[i] * order.premium / order.principal; // TODO guard order of operation?
+      returned += a[i] * (order.premium / order.principal);
```
robrobbins, line 117: long as you know this is right. FTR we should always use parens to make OO clear. Particularly on BC as I'm pretty damn sure PEMDAS is _not_ observed. 

this is good then, as long as its the correct thing.

test/lender/Lender.sol
```diff
@@ -102,31 +102,34 @@ contract Lender {
...
     ISwivel(swivelAddr).initiate(o, a, s);
 
-    // TODO: lend the remaining amount to yield? (this is prob wrong)
-    Safe.transfer(IErc20(u), y, returned - lent);
+    // lend the remaing amount to the given yield pool
+    lend(p, u, m, y, Cast.u128(returned));
 
-    emit Lend(p, u, m, returned);
-    return returned;
+    emit Lend(p, u, m, lent);
```
robrobbins, line 128: if `lend` is called from here, wouldn't emitting a lend cause a double emit of lend? 

i don't think we want that. 

if we do not emit here are we then missing some info we would want (looks like `lent` from _this_ event)

hmm. something to think about, we can discuss.

robrobbins (changes requested):
let's discuss this double emit sitch...

robrobbins (approved):

PR #246, opened 2022-05-18
Verify redeem and lend functions for pendle

sourabhmarathe:


test/lender/Lender.sol
```diff
@@ -174,11 +174,12 @@ contract Lender {
   /// @param u address of an underlying asset
   /// @param m maturity (timestamp) of the market
   /// @param a the amount of underlying tokens to lend
-  /// @param mb the minimum amount of zero-coupon tokens to return accounting for slippage
+  /// @param minimumBought the minimum amount of zero-coupon tokens to return accounting for slippage
   /// @param d the maximum timestamp at which the transaction can be executed
-  function lend(uint8 p, address u, uint256 m, uint256 a, uint256 mb, uint256 d) public returns (uint256) {
+  function lend(uint8 p, address u, uint256 m, uint256 a, uint256 minimumBought, uint256 d) public returns (uint256) {
```
robrobbins: can `minimumBought` just be `r` ?

don't confuse our bitching about Element's `return_` var with this.

it **IS** the min amount of ZCT to return right? i mean the comment says so.. i like `r`

test/redeemer/Redeemer.sol
```diff
@@ -92,18 +92,19 @@ contract Redeemer {
...
+    // Get the balance of tokens to be redeemed by the user
     uint256 amount = token.balanceOf(illuminate);
 
+    // Transfer the user's tokens to the redeem contract
     Safe.transferFrom(token, illuminate, address(this), amount);
 
+    // Redeem the tokens from the pendle contract
     IPendle(pendleAddr).redeemAfterExpiry(i, u, m);
-
-    emit Redeem(p, u, m, amount);
```
robrobbins, line 105: why does this `redeem` not emit the event?

robrobbins (changes requested):

test/lender/Lender.sol
sourabhmarathe: Yes it is, this can be `r`.

test/redeemer/Redeemer.sol
sourabhmarathe: That should not have been removed.

robrobbins (approved):
looks good. 

make sure if we have that whole `minBought` vs `r` anywhere else to unify them all. i remember that in a few places between `lender` and `redeemer`

PR #248, opened 2022-05-18
Verify redeem and lend methods for sense

sourabhmarathe:
Just one more to go...


test/lender/Lender.sol
```diff
@@ -244,35 +244,42 @@ contract Lender {
...
-    /// @notice Can be called before maturity to lend to APWine while minting Illuminate tokens
+  /// @notice Can be called before maturity to lend to APWine while minting Illuminate tokens
   /// @param u the underlying token being redeemed
   /// @param m the maturity of the market being redeemed
   /// @param a the amount of underlying tokens to lend
   /// @param ma the minimum amount of zero-coupon tokens to return accounting for slippage
   /// @param po the address of a given APWine pool
   /// @param i the id of the pool
+  /// @return returned the amount of underlying tokens that were lent
   function lend(uint8 p, address u, uint256 m, uint256 a, uint256 ma, address po, uint256 i) public returns (uint256) {
```
robrobbins: `ma` should be `r` same as elsewhere

robrobbins: also, at one point didn't we have `r` as consistently the last arg in functions? should `i` which is a pool id be last here? aren't there other functions where a pool id is present and is not the last arg?

TL;DR not only the names (ish) but the cadence of the args to a method should be consistent when possible

test/redeemer/Redeemer.sol
```diff
@@ -119,15 +122,19 @@ contract Redeemer {
   /// @param p value of a specific principal according to the Illuminate Principals Enum
   /// @param u underlying token being redeemed
   /// @param m maturity of the market being redeemed
-  /// @param d sense wut (divider?) ?
-  /// @param a sense wut (adapter?) ?
+  /// @param d sense contract that splits the loan's prinicpal and yield
+  /// @param a sense adapter contract that controls the underlying token
   function redeem(uint8 p, address u, uint256 m, address d, address a) public returns (bool) {
```
robrobbins: oof `adapter`.... have we officially adopted that term into our language? wasn't it kind a julian invention?

or, does the principal use that in their language? i'd think it'd be better to have a clearer explaination like `d` has

test/lender/Lender.sol
robrobbins: `po` should be a full word if there's no single letter for it

test/redeemer/Redeemer.sol
```diff
@@ -119,15 +122,19 @@ contract Redeemer {
...
+    // Get the principal token for the given market
     IErc20 token = IErc20(IIlluminate(illuminate).markets(u, m)[p]);
 
+    // Get the balance of tokens to be redeemed by the user
     uint256 amount = token.balanceOf(address(this));
```
robrobbins: prob save `address(this` as `self` to avoid doing the operation multiple times?

robrobbins (changes requested):

robrobbins (approved):
ship it. we'll prob arrive at exact terms for some of these relationships as we go

hopefully its something we can be _really_ general about as we gain more understanding

PR #251, opened 2022-05-19
implement build steps. update makefile

robrobbins:


test/tokens/Underlying.sol
```diff
@@ -1,22 +0,0 @@
...
-  mapping (address => mapping (address => uint256)) public allowances;
-
-  function balanceOf(address o) public view returns (uint256) {
-    return balances[o]; // this should not matter to the abi. i think...
-  }
-
-  function allowance(address o, address s) public view returns (uint256) {
-    return allowances[o][s];
-  }
-}
```
sourabhmarathe, line 22: Looks like this entire file was deleted and could be removed?

test/tokens/PErc20.sol
```diff
@@ -2,10 +2,10 @@
 
 pragma solidity 0.8.13;
 
-import "./IPErc20.sol";
+import './IPErc20.sol';
 
 /**
- * @dev Implementation of the {IERC20} interface.
+ * @dev Implementation of the {ERC20} interface, with additions
```
sourabhmarathe, line 8: I think we can just say that this is an implementation of a permissioned ERC20 token.
```suggestion
 * @dev Implementation of the permissioned {ERC20} interface
```

build/lender/Swivel.sol
```diff
@@ -0,0 +1,24 @@
+// SPDX-License-Identifier: UNLICENSED
```
sourabhmarathe, line 1: Not related to this PR, but a concern I have here is that in the future, if we wanted to implement Swivel specific functionality for the lender, we'd have a file name conflict here. Should this file be renamed to something like `Types.sol`?

build/illuminate/IPErc20.sol
```diff
@@ -0,0 +1,80 @@
+// SPDX-License-Identifier: UNLICENSED
+
+pragma solidity 0.8.13;
+
+/**
```
sourabhmarathe, line 5: We typically don't include docs for methods in interfaces -- should these also be removed here?

build/illuminate/IErc2612.sol
```diff
@@ -0,0 +1,56 @@
...
+ * @dev Interface of the ERC2612 standard as defined in the EIP.
+ *
+ * Adds the {permit} method, which can be used to change one's
+ * {IERC20-allowance} without having to send a transaction, by signing a
+ * message. This allows users to spend tokens without having to hold Ether.
+ *
+ * See https://eips.ethereum.org/EIPS/eip-2612.
+ */
+interface IErc2612 {
+    /**
```
sourabhmarathe, line 15: Same here - doesn't match our usual interface style with the docs for abstract methods.

build/illuminate/Erc2612.sol
```diff
@@ -0,0 +1,52 @@
...
+  */
+  function permit(address o, address spender, uint256 a, uint256 d, uint8 v, bytes32 r, bytes32 s) public virtual override {
+    require(d >= block.timestamp, 'erc2612 expired deadline');
+
+    bytes32 hashStruct = Hash.permit(o, spender, a, nonces[o]++, d);
+    bytes32 hash = Hash.message(domain, hashStruct); 
+    address signer = ecrecover(hash, v, r, s);
+
+    require(signer != address(0) && signer == o, 'erc2612 invalid signature');
+    _approve(o, spender, a);
```
sourabhmarathe, line 50: I don't see where this method is defined -- is it implemented somewhere else?

sourabhmarathe (changes requested):
Should be good to go, just a few questions.

test/tokens/Underlying.sol
robrobbins: it is removed? where is it showing up?

robrobbins: image

build/lender/Swivel.sol
robrobbins: `types` fits our go lang pattern and the go paradigms overall

.sol is a whole diff beast. 

by having the Principal specific constructs in a file by their name we can refer to them by

`Swivel.Order` and `Element.SwapWhatever`

robrobbins: also any argument that begins with `..in the future` i almost immediately stop reading...

lol..

test/tokens/PErc20.sol
robrobbins: yeah, but it isn't. the "permissioned interface" is IERC2612, this is a dependency for it

this whole chain of dependencies is a victim of O.Z. stupidity and I don't want to rewrite the whole thing as auditors like the OZ bs. This thing actually is just an expanded ERC20 - it has their specific additions to the "standard".

its named PErc20 to denote that it's the basis for the 'permissioned' ERC2612, as the next line of comments explain

it could have been OZERC  or EERC but we used PErc20 and its gunna stay that way, OZs dependency chain is super-fucked and i'm not opening that box again. if we rename the PErc20 it'd be a mismatch with swivel and we ain't goin there

build/illuminate/Erc2612.sol
robrobbins: this is all OZ bullshit. it's in the PErc20.

build/illuminate/IErc2612.sol
robrobbins: cuz its OZ code.

build/illuminate/IPErc20.sol
robrobbins: it's OZ code. i modify these as little as possible.

robrobbins:
no changes to make.

comments about comments are generally about OZ code which we will leave as is.

that Underlying file **is** deleted right? i don't have it in my tree at all.

PErc20 is a bit of a goofy name, but finding a sane place to work in the OZ dependency chain is impossible. there were, like, 14 fucking files i aggregated to that one. It's the dependency for the Permissioned ERC2612. A bridge between Erc20 and Erc2612.

sourabhmarathe (approved):

PR #254, opened 2022-05-20
Set illuminate in setter for redeemer and lender

### LINKED ISSUE #253: Update constructors to reflect deployment order
We will be deploying the `lender` and `redeemer` contracts prior to deploying the `illuminate` (marketplace) contract.

This means we need to remove any references to illuminate from the constructors of the `lender` and `redeemer` contracts, and add a `setIlluminate` call to those methods. This method will likely need to be `authorized`/`onlyOwner`, but that can be saved for a future ticket.

sourabhmarathe:
This PR closes #253.


pkg/lendertesting/lend_test.go
```diff
@@ -179,6 +179,9 @@ func (s *lendTestSuite) SetupSuite() {
 		},
 	}
 
+	s.Lender.SetIlluminate(s.Dep.IlluminateAddress)
```
robrobbins: see method name comment.

build/lender/Lender.sol
```diff
@@ -21,15 +21,21 @@ contract Lender {
...
+  /// @param t: the tempus contract
+  constructor(address s, address p, address t) {
     admin = msg.sender;
-    illuminate = i; // TODO add an authorized setter for this?
     swivelAddr = s;
     pendleAddr = p;
     tempusAddr = t;
   }
 
+  function setIlluminate(address i) external {
```
robrobbins: let's follow the Swivel pattern and use `..Address` in the name

https://github.com/Swivel-Finance/swivel/blob/main/contracts/v2/marketplace/MarketPlace.sol#L39=

so `setIlluminateAddress`

also

follow the swivel pattern and require the address to never have been set. "mutable once" kind of "immutable" ...

also

this needs to be `authorized` -- you can do the auth modifier stuff in another PR of course.

robrobbins (changes requested):

robrobbins (approved):

PR #259, opened 2022-05-24
Introduce feenominator attribute to lender 

### LINKED ISSUE #257: Add `feenominator` attribute to the lender contract
In Swivel, we currently set the fee using a list of `uint256`. We provide a setter for that method and that allows us to control the fee rate on initiating and exiting positions. We can implement something similar for illuminate by adding a `feenominator` attribute to the `Lender.sol` contract that will extract some fee based on the value of its `feenominator`. 

We will also implement a `setFeeRate` method that must be `authorized` by the `admin`.

sourabhmarathe:
This PR is the start of adding fee extraction to the lender contract. It simply adds a `feenominator` attribute which is similar to the list of `uint256` `feenominators` in `Swivel.sol`. Because Illuminate only has one type of lend, we only need a single `feenominator` for the lender. 

Because we don't know a ton from product yet about how specifically we want to extract fees, this PR will just get the ball rolling on that front.

In the future, if we wish to add fees upon exiting, we can add a `feenominator` to the redeemer as well.


build/lender/Lender.sol
```diff
@@ -18,19 +18,27 @@ contract Lender {
...
   /// @param p: the pendle contract
   /// @param t: the tempus contract
-  constructor(address s, address p, address t) {
+  /// @param f: set to be the feenominator, is correlated to the fee rate
+  constructor(address s, address p, address t, uint256 f) {
     admin = msg.sender;
     swivelAddr = s;
     pendleAddr = p;
     tempusAddr = t;
+    feenominator = f;
```
robrobbins: so `feenominator` was used in swivel because it was always the denominator in the calculation of the fee. if that's the case here then the name is good obvs.

build/lender/Lender.sol
```diff
@@ -18,19 +18,27 @@ contract Lender {
...
   }
 
+  function setFee(uint256 f) external authorized(admin) {
```
robrobbins, line 37: `setFee` is ok, we use it in swivel vs `setFeenominator` so... :shrug:

test/lender/Lender.sol
```diff
@@ -18,17 +18,25 @@ contract Lender {
...
-  constructor(address s, address p, address t) {
+  /// @param f: set to be the feenominator, is correlated to the fee rate
+  constructor(address s, address p, address t, uint256 f) {
     admin = msg.sender;
     swivelAddr = s;
     pendleAddr = p;
     tempusAddr = t;
+    feenominator = f;
+  }
+
```
robrobbins, line 38: stop leaving out the dev comments yo

robrobbins (changes requested):

test/lender/Lender.sol
robrobbins: still no param comment

robrobbins (changes requested):

test/lender/Lender.sol
```diff
@@ -26,6 +26,8 @@ contract Lender {
   /// @param s: the swivel contract
   /// @param p: the pendle contract
   /// @param t: the tempus contract
+  /// @dev the constrcutor sets a default value for the feenominator, which
```
robrobbins, line 29: the pattern in swivel was that @dev went above @param  (and @notice above that)

    @notice ...
    @dev ...
    @param ...
    @param...

test/lender/Lender.sol
```diff
@@ -34,8 +36,12 @@ contract Lender {
     feenominator = 1000;
   }
 
-  function setFee(uint256 f) external authorized(admin) {
+  /// Sets the feenominator to the given value
```
robrobbins, line 39: likely @notice? idr all the natspec -- take a look at swivel maybe...

robrobbins (approved):
2 comments you can address in other PRs

nothing to stop this one

PR #272, opened 2022-05-25
Extract fees from all protocols

sourabhmarathe:
This pattern was very basic, so I chose to do all of them in one PR.

This addresses the following issues: #262 #263 #264 #265 #266 #267


test/lender/Lender.sol
```diff
@@ -127,21 +124,29 @@ contract Lender {
...
+        // iterate through each order a calculate the total lent and returned
+        for (uint256 i = 0; i < o.length; i++) {
+          Swivel.Order memory order = o[i];
+          // Require the Swivel order provided matches the underlying and maturity market provided    
+          require(order.maturity == m, 'swivel maturity != maturity');
+          require(order.underlying == u, 'swivel underlying != underlying');
+          // Determine the fee
+          uint256 fee = a[i] / feenominator;
+          // Sum the total fees for to be stored in the contract
+          fees += fee;
```
sourabhmarathe: `fees` is unused, remove.

test/lender/Lender.sol
```diff
@@ -256,13 +253,15 @@ contract Lender {
...
+      // Get the underlying token
       IErc20 underlyingToken = IErc20(u);
+
+      // Transfer funds from user to Illuminate, Scope to avoid stack limit
       Safe.transferFrom(underlyingToken, msg.sender, address(this), a);
 
       // Swap on the Tempus Router using the provided market and params
       IZcToken illuminateToken = IZcToken(IIlluminate(illuminate).markets(u, m)[uint256(Illuminate.Principals.Illuminate)]);
-      uint256 returned = ITempus(tempusAddr).depositAndFix(Any(x), Any(t), a, true, r, d) - illuminateToken.balanceOf(address(this));
+      uint256 returned = ITempus(tempusAddr).depositAndFix(Any(x), Any(t), a - a / feenominator, true, r, d) - illuminateToken.balanceOf(address(this));
```
sourabhmarathe, line 264: Let's scope these calls so that we can maintain the same style throughout. This will likely get optimized out later on, but it's going to be good to have while we test out the contracts.

build/lender/Lender.sol
```diff
@@ -127,21 +124,25 @@ contract Lender {
...
+        // iterate through each order a calculate the total lent and returned
+        for (uint256 i = 0; i < o.length; i++) {
+          Swivel.Order memory order = o[i];
+          // Require the Swivel order provided matches the underlying and maturity market provided    
+          require(order.maturity == m, 'swivel maturity != maturity');
+          require(order.underlying == u, 'swivel underlying != underlying');
+          // Determine the fee
+          uint256 fee = a[i] / feenominator;
+          // Sum the total amount lent to Swivel (amount of zc tokens to mint) minus fees
+          lent += a[i] - fee;
```
robrobbins, line 137: looks like there are specs on this `lent` amount. :heavy_check_mark: if so.

build/lender/Lender.sol
```diff
@@ -97,8 +91,11 @@ contract Lender {
 
     // transfer from user to illuminate
     Safe.transferFrom(IErc20(u), msg.sender, address(this), a);
+
+    // Determine the fee
+    uint256 fee = a / feenominator;
     
-    uint256 returned = yield(u, y, a);
+    uint256 returned = yield(u, y, a - fee);
```
robrobbins, line 98: as @JTraversa could tell you, in a later audit ppl will undoubtedly bring up that an `a` smaller than fee would result in a revert and that it should be guarded by a require... we have traditionally said nonsense, just let if fail there then vs writing even more code to have it `fail better` one line earlier.

obvs no changes to be made here, but worth thinking about as you really dig into the code

build/lender/Lender.sol
```diff
@@ -127,21 +124,25 @@ contract Lender {
...
+          Swivel.Order memory order = o[i];
+          // Require the Swivel order provided matches the underlying and maturity market provided    
+          require(order.maturity == m, 'swivel maturity != maturity');
+          require(order.underlying == u, 'swivel underlying != underlying');
+          // Determine the fee
+          uint256 fee = a[i] / feenominator;
+          // Sum the total amount lent to Swivel (amount of zc tokens to mint) minus fees
+          lent += a[i] - fee;
+          // Sum the total amount of premium paid from Swivel (amount of underlying to lend to yield)
+          returned += a[i] * (order.premium / order.principal);
```
robrobbins: same for this amount, we call the `yield` internal with it so make sure its got a spec on it like `lent`, i'll double check -- ignore this if so.

build/lender/Lender.sol
```diff
@@ -222,13 +220,12 @@ contract Lender {
...
       Safe.transferFrom(IErc20(u), msg.sender, address(this), a);
 
-
       address[] memory path = new address[](2);
       path[0] = u;
       path[1] = principal;
 
       // Swap on the Pendle Router using the provided market and params
-      uint256 returned = IPendle(pendleAddr).swapExactTokensForTokens(a, r, path, address(this), d)[0];
+      uint256 returned = IPendle(pendleAddr).swapExactTokensForTokens(a - a / feenominator, r, path, address(this), d)[0];
```
robrobbins: if you are _damn certain_ about the order of operations .sol will do. remember, we usually specify them shits... not saying you _have_ too, but we generally do

if we want to relax the OO guarding parens we should make a ticket to audit them all so we aren't being inconsistent

build/lender/Lender.sol
```diff
@@ -256,13 +253,15 @@ contract Lender {
...
+      // Get the underlying token
       IErc20 underlyingToken = IErc20(u);
+
+      // Transfer funds from user to Illuminate, Scope to avoid stack limit
       Safe.transferFrom(underlyingToken, msg.sender, address(this), a);
 
       // Swap on the Tempus Router using the provided market and params
       IZcToken illuminateToken = IZcToken(IIlluminate(illuminate).markets(u, m)[uint256(Illuminate.Principals.Illuminate)]);
-      uint256 returned = ITempus(tempusAddr).depositAndFix(Any(x), Any(t), a, true, r, d) - illuminateToken.balanceOf(address(this));
+      uint256 returned = ITempus(tempusAddr).depositAndFix(Any(x), Any(t), a - a / feenominator, true, r, d) - illuminateToken.balanceOf(address(this));
```
robrobbins: samesies

robrobbins (commented):
the earlier comments hold throughout.

1. are OO safe, consistent
2. tests on lent, returned. (and any possible other side effect of those methods with which we can test our expectations && math)

good stuff.
i think.
*

build/lender/Lender.sol
JTraversa: Yeah unnecessary validation is something we err away from + in this case `fee` is derived from `a` so it cant be, and I *think* the native safeMath prevents that underflow now anyways?

robrobbins (approved):
technically looks good.

feel free to get a :+1: from @JTraversa if the product expectations are met, which you may have already done.

build/lender/Lender.sol
sourabhmarathe: The [docs](https://docs.soliditylang.org/en/v0.8.0/types.html?highlight=min#division) indicate that overflows can occur for signed integers. And like you mentioned, since the fee is calculated based on the amount, it is not possible to compute a fee that is greater than the amount.

Issue #270, opened 2022-05-25
Implement withdraw method


sourabhmarathe:
Similar to that in [Swivel.sol](https://github.com/Swivel-Finance/swivel/blob/main/contracts/v2/swivel/Swivel.sol#L401), we need a method to withdraw fees.

PR #274, opened 2022-05-25
Implement fee withdrawl to admin address

sourabhmarathe:
This PR address #270


build/lender/Lender.sol
```diff
@@ -361,6 +382,23 @@ contract Lender {
...
+  /// @notice Allows the admin to withdraw the given token based on the fees
+  /// accumulated of that token by the lender contract
+  /// @param e Address of token to withdraw
+  /// @return true if successful
+  function withdraw(address e) external authorized(admin) returns (bool) {
+    // Get the token to be withdrawn
+    IErc20 token = IErc20(e);
+    
+    // Transfer the accumulated fees to the admin
+    Safe.transfer(token, admin, fees[e]);
```
robrobbins, line 394: this order
* transfer
* reset fees[e] to 0

is vulnerable to re-entrancy. now, this is an admin method, but we should just train ourselves to always do the withdrawal first **then** call transfer. this would mean getting a ref to the amout first, so its one extra var but correct is correct -- and i guaran-damn-tee any good auditor is gunna catch it

robrobbins: might be a good idea to make a ticket to check for any other of the same... we all do it once our brain goes numb from looking at sol code all day

robrobbins (changes requested):

robrobbins (approved):

PR #276, opened 2022-05-26
Disable fees when feenominator is set to zero

### LINKED ISSUE #271: Add some way to disable fees
See Swivel.sol.

sourabhmarathe:
This PR addresses issue #271. 

We now get the fee by calling an `internal view` method called `computeFee`. If `feenominator` is 0, then we return a fee of 0. Otherwise, we compute the fee. 

We also set a `MIN_FEENOMINATOR` which sets a maximum rate for fees. This is in keeping with the pattern set forth in [Swivel.sol](https://github.com/Swivel-Finance/swivel/blob/main/contracts/v2/swivel/Swivel.sol#L419).


build/lender/Lender.sol
```diff
@@ -41,9 +42,17 @@ contract Lender {
...
   /// @return true if successful
   function setFee(uint256 f) external authorized(admin) returns (bool) {
+    require(f >= MIN_FEENOMINATOR, 'fee too high');
     feenominator = f;
     return true;
   }
+
+  /// @notice Disables fees
+  /// @return true if successful
+  function disableFees() external authorized(admin) returns (bool) {
```
robrobbins: nah. this is not needed. 

make `setFee` able to take 0.

this obvs mean the `f >= MIN_FEENOMINATOR` would need to make an exception for 0 but thats far superior that writing another method to do the same thing.

build/lender/Lender.sol
```diff
@@ -389,16 +392,30 @@ contract Lender {
...
 
+  /// @notice This method returns the fee based on the amount passed to it. If
+  /// feenominator is 0, then there is no fee.
+  /// @param a Amount of underlying tokens to calculate the fee for
+  /// @return uint256 The total for for the given amount
+  function calculateFee(uint256 a) internal view returns (uint256) {
+    if (feenominator == 0) {
+      return 0;
+    } 
+    return a / feenominator;
```
robrobbins: solidity does have a ternary, use it

`return feenominator > 0 ? a / feenominator : 0`

also can this internal function be `pure` vs `view` ? do it if so.

test/lender/Lender.sol
```diff
@@ -41,9 +42,17 @@ contract Lender {
...
   /// @return true if successful
   function setFee(uint256 f) external authorized(admin) returns (bool) {
+    require(f >= MIN_FEENOMINATOR, 'fee too high');
     feenominator = f;
     return true;
   }
+
+  /// @notice Disables fees
+  /// @return true if successful
+  function disableFees() external authorized(admin) returns (bool) {
```
robrobbins: same, use setFee that can take 0

test/lender/Lender.sol
```diff
@@ -402,6 +405,17 @@ contract Lender {
...
 
+  /// @notice This method returns the fee based on the amount passed to it. If
+  /// feenominator is 0, then there is no fee.
+  /// @param a Amount of underlying tokens to calculate the fee for
+  /// @return uint256 The total for for the given amount
+  function calculateFee(uint256 a) internal view returns (uint256) {
+    if (feenominator == 0) {
+      return 0;
+    } 
+    return a / feenominator;
```
robrobbins: same..

robrobbins (dismissed):

build/lender/Lender.sol
sourabhmarathe: It cannot be `pure` due to the reference to `feenominator`

robrobbins (approved):

PR #290, opened 2022-06-01
Address inconsistencies between lender and redeemer contracts

sourabhmarathe:
- marketplace setter function capitalization
- documentation for constructor
- ~~`marketplace` -> `marketplaceAddr`~~


build/lender/Lender.sol
```diff
@@ -11,7 +11,7 @@ import './Cast.sol';
 
 contract Lender {
   address public admin;
-  address public marketplace;
+  address public marketplaceAddr;
```
robrobbins: so, there was this fine line about the existence of `addr` on the end vs _not on the end_. 

in swivel we used it for things not in the protocol confederation itself - which would mean that this would not have it, as `marketplace` is one of the 3 contracts in the confederation. meaning we would not use `lenderAddr` or `redeemerAddr` in the marketplace contract...

but

I'm fine with adding it if it has another consistency. like, maybe all addresses that are contracts have it.

admin, not a contract, is not _adminAddr_ for example. :shrug:

robrobbins (changes requested):
ok, so, if `marketPlace` is becoming `marketPlaceAddr` in lender, redeemer then `redeemer` needs to become `redeemerAddr` in MarketPlace.sol

that, i think, would have all addresses that are contracts with ...addr on them. 

flipside -- marketPlace does not have ...addr thus have the swivel v2 pattern.

TBC i actually like saying something like , all addresses that are contracts have ...addr on them if we are consistent - this way we reconcile the fact that `admin` is `admin` an not `adminAddr`

if we go to that i'd make that change in swivel v3 as well. 
*wallet (pub key) does not have addr
*contract does

we can discuss

robrobbins (approved):

PR #302, opened 2022-06-03
Remove marketplace interface

### LINKED ISSUE #300: Remove `IMarketPlace`
Given that Solidity generates getters for nested `mapping` attributes, we no longer need to create our own getter. Also it is more efficient to load individual `address`es as needed rather than loading an entire array each time.

sourabhmarathe:
Because Solidity provides a built-in getter for the `markets` attribute in the `Lender` and `Redeemer` contracts, we no longer need to have the `IMarketPlace` interface. This PR removes that interfaces, and updates the mock contract of the `MarketPlace.sol` to only have the `markets` attribute and a setter. 

This allows the unit tests to access a minimal `MarketPlace`.

Other changes include:
- change the type of the `marketPlace` attribute in the `Lender` and `Redeemer` contracts to be a `MarketPlace`. 
- update the unit tests to use the new set up

This PR closes #300.


build/lender/Lender.sol
```diff
@@ -50,11 +51,20 @@ contract Lender {
   /// @param m: the address of the marketplace contract
   /// @return bool true if the address was set, false otherwise
   function setMarketPlaceAddress(address m) authorized(admin) external returns (bool) {
-    require(marketPlace == address(0));
-    marketPlace = m;
+    require(address(marketPlace) == address(0));
```
robrobbins, line 54: do we have a ticket to revisit the require messages? I know we are gunna look at custom errors anyway, so maybe that doesn't matter... :shrug:

build/redeemer/Redeemer.sol
```diff
@@ -8,7 +8,7 @@ import "./Safe.sol";
 
 contract Redeemer {
   address public admin;
-  address public marketPlace;
+  MarketPlace public marketPlace;
```
robrobbins, line 11: traditionally we have stored the address and hydrated the instance on demand.

this is a bigger change, and would need its own PR likely as this would suggest swivel should do this as well (storing instances vs addresses). 

let's discuss this on Monday

robrobbins (commented):
so you've snuck in a fundamental change here that we need to discuss before proceeding

that decision (to store contract instances) would effect both illuminate and swivel.

build/lender/Lender.sol
sourabhmarathe: I just made a ticket for this. 
https://github.com/Swivel-Finance/gost/issues/310

PR #303, opened 2022-06-03
Chore/i pattern

robrobbins:
1. bump solc to ...13
2. enforce `IFoo` interface naming a la Illuminate
3. enforce `addr` and `!addr` naming convention being enforced elsewhere here, and in Illuminate


build/marketplace/MarketPlace.sol
```diff
@@ -9,8 +9,8 @@ import './VaultTracker.sol';
 contract MarketPlace {
   struct Market {
     address cTokenAddr;
-    address zcTokenAddr;
-    address vaultAddr;
+    address zcToken;
+    address vaultTracker;
```
robrobbins, line 13: it Is from a vault tracker contract instance and should have been `vaultTrackerAddr`.

regardless, fixed now to conform to the local-conferation naming convention already being enforced in swivel (and illuminate as well)

sourabhmarathe (approved):
MERGE (please rename the commit before merging)

Issue #298, opened 2022-06-03
Implement a basic router contract for illuminate AMM pools


JTraversa:
We'll be looking at integrating new YieldSpace pools, perhaps not into our repo, but they need to be tested regardless.

That said, theres the consideration for the addition for a `Router` contract.

The current yieldspace implementation we will be utilizing has a unique entry point per market. 

This is not ideal as it requires approvals on each individual market for LPs.

As an alternative we should implement a basic passthrough contract that takes `underlying` and `maturity`, looks up the pool, and passes all transfers on.

The router should be able to do a `transferFrom` from the user to the pool (i believe) and then execute with those funds already in the pool.

We can also consider something more advanced like deposits to / from compound from underlying, alongside the utilization of cTokens as the base/underlying token in the pool rather than the actual `underlying` itself.

sourabhmarathe:
This was completed as part of adding `pools` to the marketplace contract.

PR #305, opened 2022-06-04
Router + YieldSpace Pool

### LINKED ISSUE #273: Review `yield()` in the swivel lend method
When we lend with Swivel, we receive underlying and PTs.

The current implementation takes the underlying returned and purchases yield protocol PTs with it to increase yield marginally.

We should review the efficiency of this methodology, and alternatives.

### Alternative 1: Leave underlying dormant until maturity 
Pros:
- Could mint them 1:1 amounts of PTs with it
- Least gas cost

Cons: 
- Leaves potential interest on the table
--------------------
### Alternative 2: Deposit to comp and keep interest for ourselves as additional fee
Pros:
- Could mint them 1:1 amounts of PTs with it
- More $ for us

Cons:
- Doesnt optimize for user
--------------------
### Alternative 3: As is, lend out the remaining at a fixed rate
Pros:
- Utilizes all capital
- Optimizes for user yield

Cons:
- Gas costs may eclipse additional yield gain

JTraversa:
In response to #298 

## Pool

Pulled in the YieldSpace pool for us to work on here though we may want to think about its location for the audit.

I dont think that the audit really needs to cover yield's contracts, so we should probably silo it away to make sure the competing auditors dont make a mistake and audit the yieldspace stuff.

## Router
I also want us to think about how we can most easily integrate with the pools with the least amount of user approvals.

I threw together a router contract that would reduce the need for approvals on each underlying <-> maturity market pool to the one central router.

It might be possible to fold the router into the current lender so it shares approvals there, but that feels a bit weird and might be going a bit far?

There is also some technical testing likely necessary that will impact patterns. If things require an additional transferFrom per trade through the router it might not be worth it?

It depends on the accounting working as expected in yield. I'm unsure if they use the `from` in transferFrom for their internal balances when they receive transfers, or the msg.sender.

## Pool <-> Marketplace?
We should also consider where/how the pools are created. I gave marketplace the admin control a method in the router called `createPool`, and left that functionality itself in the router. We could alternatively directly create pools in the marketplace.

That then brings up the question of whether the marketplace should **be** the router? Should they even be separate?


JTraversa:
Added 5095 to the Pool/Router build to demonstrate what 5095 looks like / its integration

robrobbins (commented):
what in the name of sweet baby jesus is all this?

test/lender/Lender.sol
```diff
@@ -1,6 +1,6 @@
-// SPDX-License-Identifier: UNLICENSED
+// SPDX-License-Identifier: BUSL-1.1
 
-pragma solidity 0.8.13;
+pragma solidity >= 0.8.13;
```
sourabhmarathe, line 3: Note for future reference: we are locked on `0.8.13`.

test/router/Router.sol
```diff
@@ -0,0 +1,72 @@
...
+        SafeTransferLib.safeTransfer(ERC20(address(pool.PT())), address(pool), a);
+        return pool.buyunderlying(msg.sender, pool.buyunderlyingPreview(a), a);
+    }   
+
+    function createPool(address u, uint256 m, address i) external authorized(marketplace) returns (address p) {
+        ////////////////////////////////////////////Secs in 10 yr//sell underlying fee co-eff//sell PT co-eff
+        Pool pool = new Pool(IERC20(u), IPT(i), 23381681843, 13835058055282163712, 24595658764946068821);
+        pools[u][m] = pool;
+        return (address(pool));
+    }
```
sourabhmarathe, line 54: This should go into the marketplace contract.

test/router/Router.sol
```diff
@@ -0,0 +1,72 @@
...
+    address public immutable admin;
+    address public immutable marketplace;
+    bool public paused;
+
+    constructor(address a, address m) {
+        admin = a;
+        marketplace = m;
+    }
+
+    mapping (address => mapping (uint256 => IPool)) public pools;
```
sourabhmarathe, line 19: This will also move to the marketplace contract.

test/router/Router.sol
```diff
@@ -0,0 +1,72 @@
...
+        SafeTransferLib.safeTransfer(ERC20(address(pool.underlying())), address(pool), a);
+        return pool.sellunderlying(msg.sender, pool.sellunderlyingPreview(a));
+    }
+
+    // a = the amount of underlying bought
+    function buyUnderlying(address u, uint256 m, uint128 a) external unpaused() returns (uint128 returned) {
+        IPool pool = IPool(pools[u][m]);
+        SafeTransferLib.safeTransfer(ERC20(address(pool.PT())), address(pool), a);
+        return pool.buyunderlying(msg.sender, pool.buyunderlyingPreview(a), a);
+    }
```
sourabhmarathe, line 47: See this

test/router/Router.sol
sourabhmarathe, line 54: `createPool` will be `setPool` in the `MarketPlace.sol` and will not deploy anything. It will just update the `pools` mapping.

test/router/Router.sol
```diff
@@ -0,0 +1,72 @@
...
+
+    modifier authorized(address a) {
+        require(msg.sender == a, 'sender must be authorized');
+        _;
+    }
+
+    modifier unpaused() {
+        require(!paused, 'markets are paused');
+        _;
+    }
```
sourabhmarathe, line 71: Implement in Pools milestone.

sourabhmarathe:
Closing this PR as `YieldSpace` pools interface has been implemented in `MarketPlace.sol`. The implementation of the actual pool will be completed in a separate contract with a separate default branch as it will be considered a slightly separate project from Illuminate's contract confederation of `lender`, `redeemer` and `marketplace`.

PR #312, opened 2022-06-06
Add principal field to markets interface

### LINKED ISSUE #311: Update `IMarketPlace` markets to take in principals
Currently, the `IMarketPlace` interface does not reflect the underlying interface to the `markets` mapping that will be in the `MarketPlace.sol` contract. As a result, integration tests are unable to mint zc tokens.

sourabhmarathe:
Closes #311.

In addition, we added `balanceOf` to the mocked zc token.


build/lender/Lender.sol
```diff
@@ -454,6 +450,14 @@ contract Lender {
     return feenominator > 0 ? a / feenominator : 0;
   }
 
+  /// @notice retrieves the zc token for the given market
+  /// @param u address of the underlying
+  /// @param m uint256 representing the maturity of the market
+  /// @return address of the zc token for the market
+  function getZcToken(address u, uint256 m) internal returns (address) {
```
robrobbins: wasn't this just `zcToken` ? i generally view `get` as a garbage word.

test/marketplace/ZcToken.sol
```diff
@@ -17,6 +17,8 @@ contract ZcToken {
   string public name;
   string public symbol;
   uint8 public decimals;
+  
+  mapping (address => uint256) private balances;
```
robrobbins: this is anti-pattern, lol. you forgot. 

the pattern is to use `balanceOfReturns` see the mock Erc20
https://github.com/Swivel-Finance/gost/blob/v2/test/mocks/Erc20.sol

test/mocks/MarketPlace.sol
```diff
@@ -3,16 +3,29 @@
 pragma solidity 0.8.13;
 
 contract MarketPlace {
-    address[9] private marketsReturn;
+    struct MarketsCalledArgs {
```
robrobbins: `MarketsArgs` is pattern

test/mocks/MarketPlace.sol
```diff
@@ -3,16 +3,29 @@
...
+    mapping (address => MarketsCalledArgs) public marketsCalled;
 
-    function marketsReturns(address[9] calldata m) external {
+    function marketsReturns(address m) external {
         marketsReturn = m;
     }
 
-    function markets(address u, uint256 m) external returns (address[9] memory) {
-        marketsCalled[u] = m;
+    function marketZcTokenReturns(address z) external {
```
robrobbins: there's a method named `marketZcToken` ? wut.

robrobbins: this is just a thing on the mock then? so you can manipulate getting the ZcToken?

this name would then dictate that the variable to return would be `marketZcTokenReturn`

robrobbins (changes requested):

build/lender/Lender.sol
sourabhmarathe: No, I created this helper function to deal with the stack problem. But just `zcToken` is much cleaner.

test/mocks/MarketPlace.sol
sourabhmarathe: This is supposed to be a convenience method because a given `lend` or `redeem` call can interact with the protocols principal token and Illuminate's zc token for a given underlying and maturity. When the testing, if the 0 principal is passed, I want the mocked zc token to be returned, and otherwise we can use the protocol's return value.

sourabhmarathe: Renaming to `principalTokenReturns` and `zcTokenReturns` for clarity.

test/marketplace/ZcToken.sol
```diff
@@ -18,7 +18,7 @@ contract ZcToken {
   string public symbol;
   uint8 public decimals;
   
-  mapping (address => uint256) private balances;
+  mapping (address => uint256) private balanceOfReturn;
```
robrobbins, line 21: so this is a little diff that the simpler way its in the Erc20 mock. That one does have the limitaton of only being able to have 1 set a time to return, so if we want multiple different balances to return this is fine if that is needed.

robrobbins (approved):

PR #314, opened 2022-06-07
Check principals prior to lend and redeem function execution

### LINKED ISSUE #309: Ensure revert if interface does not match `p` provided
I believe we need to add an additional else which reverts should the p not match any of those provided in the if/elses?

sourabhmarathe:
This closes #309. This will prevent events from firing if the unintended `p` is provided to a function.

Also includes several unit testing fixes and the removal of a function that is no longer used.


build/redeemer/Redeemer.sol
```diff
@@ -90,6 +92,14 @@ contract Redeemer {
     // Get the principal token that is being redeemed by the user
     address principal = IMarketPlace(marketPlace).markets(u, m, p);
 
+    // Make sure we have the correct principal
+    if (p != uint8(MarketPlace.Principals.Swivel) &&
```
robrobbins, line 96: so we are gunna go back at some point and use custom errors for all the `require` stuff

but

why is this one in an old-school revert vs a require like `Lender` ?

robrobbins: i bet in an audit somebody is going to say -> ref `MarketPlace.Principals` as a var, then check `principals.Foo`.  ye olde refinement hoisting (js of olde...)

so, it might be word doing if checked more than once :shrug:

test/redeemer/Redeemer.sol
```diff
@@ -93,12 +96,10 @@ contract Redeemer {
...
 
     // Make sure we have the correct principal
-    if (p != uint8(MarketPlace.Principals.Swivel) &&
+    require(p != uint8(MarketPlace.Principals.Swivel) &&
         p != uint8(MarketPlace.Principals.Element) &&
         p != uint8(MarketPlace.Principals.Yield) &&
-        p != uint8(MarketPlace.Principals.Notional)) {
-        revert("Invalid principal");
-    }
+        p != uint8(MarketPlace.Principals.Notional), "Invalid principal");
```
robrobbins, line 102: don't use double quotes

robrobbins (approved):

PR #316, opened 2022-06-07
Add lender to redeemer

### LINKED ISSUE #307: Make redeemer aware of lender
I'm nearly 100% sure that fund custody lies on the lender rather than marketplace, but the current redeemer contract has all their transfers coming from marketplace rather than redeemer.

So we need to either have the lender set in the redeemers constructor as a public var, or need a setter method for it.

Then we need to change all the transferFrom(marketplace, address(this), amount) -> transferFrom(lender, address(this), amount)

### LINKED ISSUE #308: Fix redemption methods transfers
The method that covers both illuminate redemption to an external user, as well as a couple protocol redemptions to the internal redemeer contract has a bit of conflicting patterns.

In the cases where the PT redeemed  **is not** illuminate, you need to transfer the `PTs` from the lender, to the redeemer in order to allow for the following external calls to redeem them. Currently it is transferring `underlying` from the marketplace to the redeemer.

Then in the case where it **is** illuminate, we need to just simply transfer the `underlying` that already exists in the redeemer, out to the person (msg.sender) who called `redeem` and is burning their illuminate tokens. Currently it looks like its trying to transfer `underlying` from the marketplace to the redeemer.

sourabhmarathe:
See #307 and #308.


build/redeemer/Redeemer.sol
```diff
@@ -41,6 +42,12 @@ contract Redeemer {
     return true;
   }
 
+  function setLenderAddress(address l) authorized(admin) external returns (bool) {
```
robrobbins: nope. we will pass the lender into the `ctor` as we'll deploy lender first. then redeemer (passing lender addr). then marketplace (passing redeemer addr) then set marketpalece address in lender and redeemer

robrobbins (changes requested):

robrobbins (approved):

Issue #317, opened 2022-06-07
Normalize parameter args to the lend and redeem methods


sourabhmarathe:
The following pattern was violated in a few of the `lend` and `redeem` calls:
```
there seems to be a pattern established initially, which is:
- Principal
- Underlying
- Maturity
- Pool Address
- Pool Id
- Amount
- Return
- Deadline
```

sourabhmarathe:
Given that `amount` is passed to all `lend` methods, I will be going with the following ordering for arguments to the `lend` method:
```
- principal
- underlying
- maturity
- amount
- return (if exists)
- deadline (if exists)
- pool/amm address (if exists)
- pool id (if exists)
- any misc. parameters
```

sourabhmarathe:
The exception is swivel's lend method, which takes in several arrays. It will not be reordered according to the new ordering.

PR #318, opened 2022-06-07
Normalize argument order to lend

sourabhmarathe:
See #317. 

Note that we went with the following ordering:
```
- principal
- underlying
- maturity
- amount
- return (if exists)
- deadline (if exists)
- pool/amm address (if exists)
- pool id (if exists)
- any misc. parameters
```


robrobbins (approved):
imma stamp this one and assume you and @alexanderwende got to quorum on it...

PR #320, opened 2022-06-07
Rename `yToken` to `pool`

### LINKED ISSUE #313: Lender: yToken -> Pool?
https://github.com/Swivel-Finance/gost/blob/illuminate/test/lender/Lender.sol#L88

sourabhmarathe:
See #313.


sourabhmarathe:
Reasoning was the methods we were using are in the `IPool` interface provided by yield [here](https://github.com/yieldprotocol/yieldspace-interfaces/blob/main/IPool.sol).

robrobbins (approved):

PR #321, opened 2022-06-08
Implement ERC-5095 token redeem

### LINKED ISSUE #306: zcTokens -> ERC-5095
We should help lock in the final state of 5095 and then integrate it as a replacement for our zcTokens.

In Swivel this may require some minimal effort to giving some fund control over to the zcToken contract itself, allowing the zcTokens to trigger an admin method on Swivel.sol that redeems PT's.

This would mean anyone owning a zcToken could then redeem it solely knowing the address of said zcToken (which they have if they own it).

sourabhmarathe:
See #306.


build/redeemer/Interfaces.sol
```diff
@@ -19,6 +19,7 @@ interface IZcToken {
 	function mint(address, uint256) external returns (bool);
 	function balanceOf(address) external returns (uint256);
 	function burn(address, uint256) external;
+  function maturity() external view returns (uint256);
```
robrobbins: what's up with this indent?

robrobbins: also we don't put specifiers (view) on the interfaces, you can, but id doesn't buy us anything except narrowing the interface? if we wanted them we'd need to audit all the interfaces

build/redeemer/Redeemer.sol
```diff
@@ -189,6 +189,30 @@ contract Redeemer {
...
 
+
+  /// @notice implements the redeem method for the contract to fulfil the 
+  /// ERC-5095 interface
+  /// @param u address of the underlying asset
+  /// @param m maturity of the market
+  /// @param f address from where the underlying asset will be burned
+  /// @param t address to where the underlying asset will be transferred
+  /// @return bool true if the underlying asset was burned successfully
+  function redeem(address u, uint256 m, address f, address t) public authorized(admin) returns (bool) {
```
robrobbins: so, product wise, this is a different thing than the `Redeemer.redeem` exposed as public interface.

should it's name reflect what it's actually doing? I mean, I am always **fine** with being thrifty with words and names -- but this is a different thing altogether than `redeem` right? A FE / Dapp will probably want this differentiated more than just signature. My gut anyway -- i could be wrong. we can discuss.

robrobbins (changes requested):

robrobbins (commented):
lol, just make the mock for the 5095 and call that.

however,

there may be a place for this, here is my suggestion. we can actually formalize this authorization spoofing -- it comes up from time to time. auctully its just spoofing -- a way to tell the mock you need to return something else when called given the presence of that flag -- it doesn't need to know the word authorized at all...

we can stay away from `isAuthorized`, as the mock doesn't know what authorized is anyway -- we can make it understand that it's spoofing a return, thus returning something else

let's discuss

robrobbins (commented):
close, i'm thinking thru this new pattern. see comment back on discord...

robrobbins (approved):

PR #326, opened 2022-06-09
Retrieve redeemable balance from marketplace

### LINKED ISSUE #315: Check redeem of sense
Specifically this line:
`uint256 amount = token.balanceOf(self);`

Are we sure that `self` should not be `msg.sender`?

sourabhmarathe:
Sense was incorrectly retrieving its redeemable balance from the `redeemer` contract. This was not how the other protocols are doing it and is not how things are supposed to work. 

See #315 (check redeem of sense).


robrobbins (dismissed):
is this not something we have a test on?

kind of an important change with no spec changes.

ship it -- write a test if you can.

Issue #329, opened 2022-06-10
Implement pool interface and mocks


sourabhmarathe:
This means implementing `IPool` and adding a `Pool.sol` mock to the `mocks/` directory.

PR #331, opened 2022-06-10
Implement mock of YieldSpace pools

sourabhmarathe:
See #329.


build/marketplace/Interfaces.sol
```diff
@@ -5,3 +5,66 @@ pragma solidity 0.8.13;
...
+
+    function redeem(
+        uint256,
+        address,
+        address
+    ) external returns (uint256);
+}
+
+interface IPool {
+    function PT() external returns (IErc5095);
```
robrobbins, line 45: oh man i hate dealing with abbreviations... `PT` and `SellPT`. I guess that's the interface exposed by the pool tho so, it is what it is. 

generally we only capitalize the First like `Erc20` vs `ERC20` -- our rule (same in our golang code) being only capitalize what you absolute must vs some vague rule about capitalizing the first letter of whatever this abbreviates or whatever

anyway, if the pool is doing it we are stuck with it
all good

robrobbins (approved):

Issue #330, opened 2022-06-10
Implement router functionality in marketplace


sourabhmarathe:
Implement the buy/sell and liquidity provision methods to `MarketPlace.sol`. See this [part](https://github.com/Swivel-Finance/gost/pull/305/files#r894816775) of Julian's PR.

PR #333, opened 2022-06-10
Implement swaps between underlying and PTs in marketplace

sourabhmarathe:
See #330. 

This PR also includes fixes to the mocked Pool.


build/marketplace/MarketPlace.sol
```diff
@@ -22,11 +23,14 @@ contract MarketPlace {
...
-    /// markets are defined by a market pair which point to a fixed length array of principal token addresses.
-    /// the principal tokens those addresses represent correspond to their Principals enum value, thus the
-    /// array should be ordered in that way
+    /// markets are defined by a market pair which point to a fixed length array
+    /// of principal token addresses. The principal tokens those addresses
+    /// represent correspond to their Principals enum value, thus the array
+    /// should be ordered in that way
     mapping(address => mapping(uint256 => address[9])) public markets;
 
+    mapping(address => mapping(uint256 => address[9])) public pools;
```
robrobbins, line 32: oh that looks wierd at first, but yeah, i guess there are 0-8 available "slots" for a pool that is associated with an individual principal in a market

i suppose then the pool can be referenced with the Principals enum the same way a given principal (in a market) can?

build/marketplace/MarketPlace.sol
```diff
@@ -82,6 +86,91 @@ contract MarketPlace {
...
+        return true;
+    }
+
+    /// @notice sells the PT for the PT via the pool
+    /// @param p enum value of the principal token
+    /// @param u address of the underlying asset
+    /// @param m maturity (timestamp) of the market
+    /// @param a amount of PT to swap
+    /// @return uint128 amount of PT bought
+    function sellPT(
```
robrobbins: so let's think about these names on a protocol big-picture. do we want `sellPT` or `sellPrincipalToken`

in our protocols generally (including swivel) we have gone with longer expressive method names

something to discuss

also gets us away from the `PT` vs `Pt` kind of argument...

build/marketplace/MarketPlace.sol
```diff
@@ -82,6 +86,91 @@ contract MarketPlace {
...
+        return pool.buyPT(msg.sender, pool.buyPTPreview(a), a);
+    }
+
+    /// @notice sells the underlying for the PT via the pool
+    /// @param p enum value of the principal token
+    /// @param u address of the underlying asset
+    /// @param m maturity (timestamp) of the market
+    /// @param a amount of underlying to swap
+    /// @return uint128 amount of underlying sold
+    function sellUnderlying(
```
robrobbins, line 146: this isn't `sellU` for example...

robrobbins (dismissed):
let's discuss the names next week.

code looks good

build/marketplace/MarketPlace.sol
sourabhmarathe: Yes, that's what I was thinking.

robrobbins (approved):

PR #334, opened 2022-06-10
Pause pool using `unpaused` modifier

### LINKED ISSUE #207: Add per market functionality to pause()

### LINKED ISSUE #324: Add per-protocol (or per market might be better?) pause functionality to lender
If we do it per market, we should implement a batch pause with an array so that you can still pause all of a protocol at once.

sourabhmarathe:
See #331.

This PR implements a flag that pauses usage of the pools.


build/marketplace/MarketPlace.sol
```diff
@@ -34,6 +34,8 @@ contract MarketPlace {
     address public admin;
     /// @notice address of the deployed redeemer contract
     address public immutable redeemer;
+    /// @notice flag that determines if pools may be used
+    bool public paused;
```
robrobbins: this, obvs, is all markets

do we want this blanket pause or an individual way to pause a specific market?

sourabhmarathe: I'm not sure. I'll ask product.

build/marketplace/MarketPlace.sol
```diff
@@ -88,6 +88,15 @@ contract MarketPlace {
         return true;
     }
 
+    /// @notice allows the admin to pause a protocol's pools
+    /// @param p protocol's enum value
```
robrobbins: PRINCIPAL not PROTOCOL

build/marketplace/MarketPlace.sol
```diff
@@ -177,4 +186,9 @@ contract MarketPlace {
         require(msg.sender == a, 'sender must be authorized');
         _;
     }
+
+    modifier unpaused(uint8 p) {
+        require(!paused[p], 'markets are paused');
```
robrobbins: i think the err msg should be `principal is paused` or something with the fact that its Principal based?

build/marketplace/MarketPlace.sol
```diff
@@ -34,8 +34,8 @@ contract MarketPlace {
     address public admin;
     /// @notice address of the deployed redeemer contract
     address public immutable redeemer;
-    /// @notice flag that determines if pools may be used
-    bool public paused;
+    /// @notice flag that determines if a protocol's pool is available
+    bool[9] public paused = [false, false, false, false, false, false, false, false, false];
```
robrobbins, line 38: using the principals enum -- i liek it.

robrobbins (changes requested):

PR #341, opened 2022-06-13
Deposit to aave pool 

### LINKED ISSUE #340: Integrate Aave into APWine lend method
See the LendingPool contract in Aave. APWine requires that we deposit lended funds into Aave before they accept deposits. Additionally, we may need to add a mass approve function into Aave to get this functionality to work well.

sourabhmarathe:
APWine requires that we deposit user funds to Aave prior to swapping the underlying token for the fixed rate token. This PR implements that functionality.

Additionally, we will need to approve all spends to Aave, which is why we implemented an `approveUnderlying` method similar to what we already have in Swivel.

This PR closes #340.


build/lender/Lender.sol
```diff
@@ -43,6 +43,23 @@ contract Lender {
         feenominator = 1000;
     }
 
+    /// @notice Allows the admin to bulk approve given compound addresses at the underlying token, saving marginal approvals
+    /// @param u array of underlying token addresses
+    /// @param c array of compound token addresses
```
robrobbins: "compound token addresses" wut? 

we are ridding swivel of compound-specific hard coded stuff (so i doubt we wanna bring that same thing into illuminate). shouldn't `c` be something more generic here than specifically _Compound_ ?

"Compounding Token" would be a swivel-esque (v3) term...

build/lender/Lender.sol
```diff
@@ -414,6 +431,7 @@ contract Lender {
         uint256 a,
         uint256 r,
         address pool,
+        address aave,
```
robrobbins, line 436: yeah, we were bound to hit the full-word-arg at some point...

build/lender/Lender.sol
```diff
@@ -435,6 +453,10 @@ contract Lender {
         // Determine the amount lent after fees
         uint256 lent = a - fee;
 
+        // Deposit into aave
+        // TODO: Get a referral code and use it to deposit
+        IAave(aave).deposit(u, lent, address(this), 0);
+
         // Swap on the APWine Pool using the provided market and params
         uint256 returned = IAPWineRouter(pool).swapExactAmountIn(i, 1, lent, 0, r, address(this));
```
robrobbins, line 463: those hard coded values look mysterious `0` in 458 and `1` in 461.

it's prob something we are stuck with tho, not ganna make any sort of demands on this, just noting it for posterity

build/lender/Lender.sol
sourabhmarathe: We can go with that. This was directly pulled from the swivel codebase I believe.

build/lender/Lender.sol
sourabhmarathe: Yes, we are stuck with this one in particular. It's a referral code that used to be part of Aave's incentive system. I believe it is now an unused parameter (i.e. we could send anything in for that parameter). That said, the docs do recommend going with 0 for this parameter if the user doesn't have a referral code.

robrobbins: S.
M.
H.

WEB3 BRO

robrobbins (approved):

PR #353, opened 2022-06-16
Add custom errors

### LINKED ISSUE #191: Custom Error Messages
With the introduction of custom errors last year, there are some optimizations we can add.

https://blog.soliditylang.org/2021/04/21/custom-errors/

Changing our reverts to custom errors should reduce bytecode size, and slightly reduce gas costs in non-reverting cases

### LINKED ISSUE #304: Implement custom errors
We should implement custom errors for all of our current reverts, then take a quick look at them to review/audit what sort of data might be useful to bubble up in the errors themselves.

sourabhmarathe:
This PR closes #191. 

It also upgrades to geth to `v1.10.19`.


build/lender/Lender.sol
```diff
@@ -10,6 +10,12 @@ import './Safe.sol';
 import './Cast.sol';
 
 contract Lender {
+    error Unauthorized();
+    error MarketMismatch(string);
```
robrobbins: i'm a big no on "mismatch"

build/lender/Lender.sol
```diff
@@ -10,6 +10,12 @@ import './Safe.sol';
...
+    error MarketAlreadySet(address);
```
robrobbins: `Already` is a garbage word here. `MarketSet` if it needs to be narrow, or perhaps a generic `exists` or something

build/lender/Lender.sol
```diff
@@ -10,6 +10,12 @@ import './Safe.sol';
...
+    error InvalidPrincipal(uint8);
```
robrobbins: maybe if it needs to be narrow... wondering if `PrincipalInvalid` if so... idk

anyway a wide `invalid` could be useful?

we'll have to discuss how naroow / wide to be

build/lender/Lender.sol
```diff
@@ -10,6 +10,12 @@ import './Safe.sol';
...
+    error ArrayLengthMismatch();
```
robrobbins: "mismatch" ain't going in...

`ArrayLength` ,  if it need be that narrow -- we use arrays a lot so, i could see that as useful

build/lender/Lender.sol
```diff
@@ -80,7 +86,9 @@ contract Lender {
     /// @return true if successful
     function approve(address[] calldata u, address[] calldata a) external authorized(admin) returns (bool) {
         uint256 len = u.length;
-        require(len == a.length, 'array length mismatch');
```
robrobbins, line 83: oh how did i miss this word before. shame on me. well, good its being taken out.

build/lender/Lender.sol
robrobbins: somehow i missed that word before...

build/lender/Lender.sol
```diff
@@ -80,7 +86,9 @@ contract Lender {
...
+        if (len != a.length) {
+            revert ArrayLengthMismatch();
```
robrobbins: revert ArrayLength()

build/lender/Lender.sol
```diff
@@ -108,7 +116,9 @@ contract Lender {
     /// @param m: the address of the marketplace contract
     /// @return bool true if the address was set, false otherwise
     function setMarketPlaceAddress(address m) external authorized(admin) returns (bool) {
-        require(marketPlace == address(0));
+        if (marketPlace != address(0)) {
+            revert MarketAlreadySet(marketPlace);
```
robrobbins: if it stays narrow let's not use just Market, let's use its actual name so something ilke
`revert MarketPlaceSet`

if we could be wide, like other places that check for things existing 
`revert Exists` <- that may not be enuff info -- UNLESS it told you what method it reverted in, like,

`error Exists thrown in setMarketPlaceAddress` -> that is perfect, but if the info we get is only useful with a narrow name `MarketPlaceSet` kind of deal...

build/lender/Lender.sol
robrobbins: oooh, you can pass a string to these? we could go wide then

`Exists('Market')`

build/lender/Lender.sol
robrobbins: Invalid('Principal')

build/lender/Lender.sol
robrobbins: `Length('Arrays')`

build/lender/Lender.sol
robrobbins: or i like the wide ones with text if that works

build/lender/Lender.sol
robrobbins: `revert Exists('MarketPlace address')` ?

build/lender/Lender.sol
robrobbins: this is an equality check, so use that sort of language maybe?

`NotEqual('Foo, Bar')`

could be used for all kinds of shit

build/lender/Lender.sol
```diff
@@ -154,14 +164,19 @@ contract Lender {
...
+        }
 
         // uses yield token interface...
         IYield pool = IYield(y);
 
         // the yield token must match the market pair
-        require(address(pool.base()) == u, 'yield base != underlying');
-        require(pool.maturity() <= m, 'yield maturity != maturity');
+        if (address(pool.base()) != u) {
+            revert MarketMismatch('underlying');
```
robrobbins: `revert NotEqual('underlying')` ?

robrobbins (commented):
i'll stop here so we can discuss narrow/wide

wide would allow is to constrain the growth of random new terms all over the protocol

robrobbins (approved):

PR #366, opened 2022-06-20
Pause markets on lender

sourabhmarathe:
This PR also removes pausing pools in MarketPlace.
This PR also implements individual market setting (down to the principal level).

### DIFF
--- a/test/lender/Lender.sol
+++ b/test/lender/Lender.sol
@@ -26,8 +26,8 @@ contract Lender {
     address public admin;
     /// @notice address of the MarketPlace.sol contract, used to access the markets mapping
     address public marketPlace;
-    /// @notice mapping that determines if a market may be used by a lender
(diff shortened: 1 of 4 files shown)


test/lender/Lender.sol
```diff
@@ -26,6 +26,8 @@ contract Lender {
     address public admin;
     /// @notice address of the MarketPlace.sol contract, used to access the markets mapping
     address public marketPlace;
+    /// @notice mapping that determines if a market may be used by a lender
+    mapping(address => mapping(uint256 => bool)) public paused;
```
JTraversa: mb not describing better, should map the principal enum to paused rather than u,m

This lets us pause a protocol if it gets hacked, rather than a whole market

sourabhmarathe: No worries, just fixed it!

PR #370, opened 2022-06-21
feat: implement a protocol enum and use it

robrobbins:


JTraversa:
First pass, issues with custody but conceptually a good first pass 🚢  for now!

JTraversa (approved):

PR #373, opened 2022-06-22
Fix stack too deep error

sourabhmarathe:
The current lender contract was unable to compile due to the `stack too deep` error. In `0.8.13` of Solidity, this was supposed to be fixed by the update that allows moving stack variables into memory. However, this was not effective on our problem because this doesn't work for large numbers of temporary variables, which must remain on the stack. 

To get around this, I created input structures in `Input.sol`. These contain the same information, but are placed in memory instead of on the stack. Although this will cost more gas, it improves the legibility of the code and gives us enough memory in the stack to be able to perform `lend` operations as before.


JTraversa:
The current code sans-changes compiles with local scoping yeah? Or is it actually unable to compile?

Regardless im going to put this on hold until we merge all the audit changes

PR #380, opened 2022-06-22
Fix build by scoping lend methods

sourabhmarathe:
Added some brackets and etc.


JTraversa (approved):
I dont see any issues 🚢

PR #381, opened 2022-06-22
Feat/use libs

robrobbins:
* move the methods for executing TX type on Compound to library file `Compound.sol` in /swivel
* this does not impact Marketplace.sol and VaultTraker.sol as they have read-only methods addressed by ICompounding
* abstract `deposit` and `withdraw` boilerplate into one method each (these will call the library code depending on protocol)
* fix the underlying methods and specs which were still in a previous state.


robrobbins:
going with no libs instead

PR #382, opened 2022-06-22
Feat/no use libs

robrobbins:


build/swivel/Swivel.sol
```diff
@@ -111,11 +111,12 @@ contract Swivel {
 
     IMarketPlace mPlace = IMarketPlace(marketPlace);
     address cTokenAddr;
-    address adapterAddr;
-    (cTokenAddr, adapterAddr) = mPlace.cTokenAndAdapterAddress(o.protocol, o.underlying, o.maturity);
+    // TODO if libraries work, we can revert this to only cTokenAddress if desired...
+    (cTokenAddr,) = mPlace.cTokenAndAdapterAddress(o.protocol, o.underlying, o.maturity);
+
+    // perform the actual deposit type transaction, specific to a protocol
+    require(deposit(o.protocol, cTokenAddr, principalFilled), 'deposit failed');
```
sourabhmarathe, line 118: Perhaps this is early optimization, but should we consider using `revert` and implementing our own errors from the get go?

sourabhmarathe (approved):
LGTM

build/swivel/Swivel.sol
JTraversa: Should do all of them at once so we dont miss anything / end up against pattern (We said this on illuminate and i think one slipped through still)

JTraversa:
Likewise LGTM, we'll jam on the read stuff in discord

JTraversa (approved):

PR #383, opened 2022-06-23
implement, test the Erc4626 interface for transactions

robrobbins:
* create Erc4626 interface and use as Protocols[0]
* create mock
* Implement in a test to assure swivel contract calls as expected when protocol is 0


pkg/swiveltesting/dep.go
```diff
@@ -62,6 +64,14 @@ func Deploy(e *Env) (*Dep, error) {
 
 	e.Blockchain.Commit()
 
+	tvAddress, _, tvContract, tvErr := mocks.DeployErc4626(e.Owner.Opts, e.Blockchain)
+
+	if tvErr != nil {
+		return nil, tvErr
+	}
```
sourabhmarathe:
```suggestion
	erc4626Address, _, erc4626Contract, erc4626Err := mocks.DeployErc4626(e.Owner.Opts, e.Blockchain)

	if erc4626Err != nil {
		return nil, erc4626Err
	}
```
This appears a bit out of pattern. In fact, I think we could also argue that the name for the mocked ERC20 token is not extensible either -- perhaps that should be renamed to `erc20...` so that we have a convention for whenever we want to mock a token standard. Adding suggested changes to follow what we have for `erc20` tokens.

This may also be useful down the road as we deal with ERC5095.

pkg/swiveltesting/dep.go
```diff
@@ -86,10 +96,12 @@ func Deploy(e *Env) (*Dep, error) {
...
 		HashFakeAddress:      hashAddress,
 		HashFake:             hashContract,
-		Erc20Address:         ercAddress,
 		Erc20:                ercContract,
-		CompoundTokenAddress: ctAddress,
+		Erc20Address:         ercAddress,
 		CompoundToken:        ctContract,
+		CompoundTokenAddress: ctAddress,
+		Erc4626:              tvContract,
+		Erc4626Address:       tvAddress,
```
sourabhmarathe:
```suggestion
		Erc4626:              erc4626Contract,
		Erc4626Address:       erc4626Address,
```
See comment above about matching pattern.

test/mocks/Erc4626.sol
```diff
@@ -0,0 +1,45 @@
+// SPDX-License-Identifier: UNLICENSED
+
+/**
+  Erc4626 is a mock tokenized vault with stubs of the methods we need for testing.
+*/
+
+pragma solidity 0.8.13;
+
+// TODO this could inherit from the ERC20 mock if needed
+contract Erc4626 {
```
sourabhmarathe, line 10: Has this not been done because it is not needed yet? Should be pretty easy to extend off of the mocked ERC20 token contract. (This one is totally up to you)
```suggestion
// TODO this could inherit from the ERC20 mock if needed
contract Erc4626 is Erc20 {
```

sourabhmarathe (dismissed):
A couple small nits, but overall looks good!

test/mocks/Erc4626.sol
robrobbins: not needed. too much information, even if correct, is not good. with any code, but **especially** with interfaces and their implementations should never do anything past the exact minimum to succeed

pkg/swiveltesting/dep.go
robrobbins: these are temp vars and are completely insignificant.

that being said, yeah -- we can use erc20 and erc4626.

PR #384, opened 2022-06-24
implement and test yearn interface for swivel transactions:w

robrobbins:
* create, implement Yearn interface for swivel
* mock YearnVault and deploy to test


build/swivel/Interfaces.sol
```diff
@@ -14,12 +14,16 @@ interface IErc4626 {
   function withdraw(uint256, address, address) external returns (uint256);
 }
 
-// TODO if we move away from adapters the `..Token` suffix can be dropped
-interface ICompoundToken {
```
robrobbins, line 18: there's no reason to be overly specific with the interface names in Swivel.sol. These interfaces will be fine simply being named for what is in the Protocol Enum

pkg/swiveltesting/dep.go
```diff
@@ -21,6 +21,8 @@ type Dep struct {
 	CompoundTokenAddress common.Address
 	Erc4626              *mocks.Erc4626
 	Erc4626Address       common.Address
+	YearnVault           *mocks.YearnVault
+	YearnVaultAddress    common.Address
```
robrobbins, line 25: the names are a bit more specific in mocks and tests as they need to not clash with mock adapters. these names don't effect the actual protocol at all, as, they are just here for testing

test/swivel/Swivel.sol
```diff
@@ -583,8 +587,12 @@ contract Swivel {
...
   /// @param a Amount to withdraw
   function withdraw(uint8 p, address c, uint256 a) internal returns (bool) {
-    if (p == uint8(Protocols.Compound)) {
-      return ICompoundToken(c).redeemUnderlying(a) == 0;
+    // TODO as stated elsewhere, we may choose to simply return true here and not attempt to measure against any expected return
+    if (p == uint8(Protocols.Compound)) { // TODO is Rari a drop in here?
+      return ICompound(c).redeemUnderlying(a) == 0;
+    } else if (p == uint8(Protocols.Yearn)) {
+      // yearn vault api states that withdraw returns uint256
+      return IYearn(c).withdraw(a) >= 0;
```
sourabhmarathe, line 595: Early optimization, but an easy one. `uint256`s tested for `>=` to 0 can be more efficiently checked via the `!=` operator.
```suggestion
    } else if (p == uint8(Protocols.Yearn)) {
      // yearn vault api states that withdraw returns uint256
      return IYearn(c).withdraw(a) != 0;
```

sourabhmarathe (approved):
One early optimization. Looks good!

test/swivel/Swivel.sol
robrobbins: 0 is a `uint256` allowable value and I don't know that `0` would never be returned. Thus `!= 0` may revert on an acceptable value.

PR #387, opened 2022-06-28
feat: implement and test Euler interface for swivel

robrobbins:


Makefile
```diff
@@ -35,7 +36,7 @@
 
 .PHONY: all
 
-# Mocks
+# ------------------------------------------------------- MOCKS -------------------------------------------------------------
```
robrobbins, line 39: after we ship illuminate and v3 @sourabhmarathe we'll sit down and rework these growing makefiles
things can be made less append-for-the-win by using some variables and some better composition....

this works til then

JTraversa (approved):
Pretty straightforward

PR #392, opened 2022-07-05
Feat/compounding euler

robrobbins:
* Euler implemented in Compounding Library
* Test Marketplace
* Test VaultTracker

TODO: check that `convertBalanceToUnderlying(1e27)` is correct for "exchangeRate"


JTraversa:
Should probably just make both the consts either 1e27 or 1e26, just to keep things consistent if theyre both arbitrary

JTraversa (approved):
Other than that lgtm

PR #393, opened 2022-07-07
Feat/custom errors swivel

robrobbins:
custom errors for Swivel.sol

this branch was _going_ to be vault tracker , but i changed my mind so ignore the shitty branch name. Its primarily Swivel.sol and its specs (with only skeletal introduction to the other contracts yet - going to do them one at a time)

see the errors.txt files for synopsis. we can remove the TODO's once each have been reviewed

much like with the libraries i'm going to PR the other branches against THIS ONE first then merge it all to v3 if they will all compile.


sourabhmarathe (dismissed):
LGTM

PR #394, opened 2022-07-08
feat: custom errors in Marketplace.sol

robrobbins:


sourabhmarathe (approved):

build/marketplace/MarketPlace.sol
```diff
@@ -67,11 +72,20 @@ contract MarketPlace {
     string memory n,
     string memory s
   ) external authorized(admin) unpaused() returns (bool) {
-    require(swivel != address(0), 'swivel contract address not set');
+    // TODO remove on confirm
+    // require(swivel != address(0), 'swivel contract address not set');
+    if (swivel == address(0)) {
+      revert Exception(21, 0, 0, address(0), address(0));
```
JTraversa, line 78: Might be able to get away with a generic "address not set" if there are other occurrences in Swivel.sol itself?

build/marketplace/MarketPlace.sol
```diff
@@ -96,15 +110,28 @@ contract MarketPlace {
...
+    // require(block.timestamp >= m, 'maturity not reached');
+    if (block.timestamp < m) {
+      revert Exception(24, block.timestamp, m, address(0), address(0));
+    }
 
     // set the base maturity cToken exchange rate at maturity to the current cToken exchange rate
     uint256 exchangeRate = Compounding.exchangeRate(p, market.cTokenAddr);
     markets[p][u][m].maturityRate = exchangeRate;
 
     // TODO remove the require here? why?
```
JTraversa: I think just remove it imo.

robrobbins: done

build/marketplace/MarketPlace.sol
robrobbins: there's not any others, it's specific to the need of having swivel's address set.

plus, the list is @alexanderwende approved so ...

PR #404, opened 2022-07-14
bug: add maturity gate for zcToken Mint and specs

robrobbins:


JTraversa (dismissed):
The zcToken looks good, didnt check *every* file but looks like its all a pragma bump

test/marketplace/Interfaces.sol
```diff
@@ -1,6 +1,6 @@
 // SPDX-License-Identifier: UNLICENSED
 
-pragma solidity 0.8.4;
+pragma solidity >= 0.8.4;
```
sourabhmarathe, line 3: Why aren't we locking in a Solidity version anymore?

test/tokens/ZcToken.sol
```diff
@@ -33,11 +33,13 @@ contract ZcToken is Erc2612, IZcToken {
       return true;
   }
 
+  /// @dev Muturity must not have been reached else we revert here.
   /// @param t Address recieving the minted amount
   /// @param a The amount to mint
   function mint(address t, uint256 a) external onlyAdmin(admin) override returns(bool) {
-      _mint(t, a);
-      return true;
+    require(block.timestamp < maturity, 'maturity reached');
```
sourabhmarathe, line 40: Use a custom error here instead and revert to save gas.

sourabhmarathe (changes requested):
A question and a small gas saving suggestion.

test/tokens/ZcToken.sol
robrobbins: not introducing custom errors in V2.

test/marketplace/Interfaces.sol
robrobbins: while i prefer locking to a specific version it seems that @JTraversa and everyone else in the space is fine with this. It's all over the v3 repo.

sourabhmarathe (dismissed):

PR #405, opened 2022-07-15
implement v2 fee fixes, features and specs

robrobbins:


pkg/swiveltesting/exit_zctoken_filling_zctoken_initiate_test.go
```diff
@@ -166,17 +166,16 @@ func (s *EZFZISuite) TestEZFZI() {
 	assert.Equal(amt, amount)
 
 	// first call to utoken transferfrom 'from' should be maker here...
```
robrobbins, line 168: this block had been commented out because of how the mocks will overwrite successive calls to the same method if the calling address is the same, with the fix in EZFZI (the msg.sender change) it allows us to check both of these sequentially now as the `transferFrom` is called with .Maker, then msg.sender (Owner.Opts.From in the test)

that is why this is uncommented now -- it occurs to me that this was actually catching the bug and i didn't realize it.

JTraversa (dismissed):
lgtm

test/swivel/Swivel.sol
```diff
@@ -505,6 +505,20 @@ contract Swivel {
...
+  /// @notice Allows Swivel to redeem any currently accrued interest (via MarketPlace)
+  /// @param u Underlying token address associated with the market
+  /// @param m Maturity timestamp of the market
+  function redeemSwivelVaultInterest(address u, uint256 m) external returns (bool) {
+    MarketPlace mPlace = MarketPlace(marketPlace);
+    // call marketplace to determine the amount redeemed
+    uint256 redeemed = mPlace.redeemVaultInterest(u, m, address(this));
+    // redeem underlying from compound
+    require(CErc20(mPlace.cTokenAddress(u, m)).redeemUnderlying(redeemed) == 0, 'compound redemption failed');
+    // NOTE: for swivel reddem there is no transfer out as there is in redeemVaultInterest
```
sourabhmarathe:
```suggestion
    // NOTE: for swivel redeem there is no transfer out as there is in redeemVaultInterest
```
typo fix

sourabhmarathe (dismissed):
LGTM, suggestion is just a typo fix

PR #409, opened 2022-07-28
preparing for issue 180 bugfix: create ZcToken Unit test suite

robrobbins:
* actual bugfix is not the biggest issue here. There were no previous unit tests on the ZcToken as it was mostly a pass thru to existing tokens. 
* much logic, and a found audit bug dictates that this contract now be setup for proper unit tests
* a new testing package is created with its accompanying setup
* one introductory spec -- the zctoken_construction_test is present to check that the token hydrates passed values correctly
* with this in place the actual spec for the the bugfix to issue 180 can be tested now
* that test is forthcoming


build/creator/IRedeemer.sol
```diff
@@ -11,5 +11,5 @@ interface IRedeemer {
 
     function authRedeem(uint8, address, uint256, address, address, uint256) external returns (uint256);
     function markets(uint8, address, uint256) external view returns (Market memory market);
-    function getExchangeRate(uint8, address) external view returns (uint256);
```
robrobbins, line 14: we don't historically use `getFoo` anywhere so i changed this to match existing patterns

build/creator/ZcToken.sol
```diff
@@ -17,7 +17,7 @@ contract ZcToken is Erc20, IERC5095 {
     /// @dev address of a cToken
     address public immutable cToken;
     /// @dev address and interface for an external custody contract (necessary for some project's backwards compatability)
-    IRedeemer public immutable redeemer;
+    address public immutable redeemer;
```
robrobbins, line 20: this contract was already expecting this to be an address and uses the `Interface(address).method()` pattern.

this is likely addressing something in the audit somewhere.

build/creator/ZcToken.sol
```diff
@@ -41,17 +42,22 @@ contract ZcToken is Erc20, IERC5095 {
         if (block.timestamp < maturity) {
             return 0;
         }
-        return (principalAmount * IRedeemer(redeemer).getExchangeRate(protocol, cToken) / IRedeemer(redeemer).markets(protocol, underlying, maturity).maturityRate);
+
+        // NOTE we weren't even using the pre-hydrated redeemer in most of these anyway...
+        return (principalAmount * IRedeemer(redeemer).exchangeRate(protocol, cToken) / IRedeemer(redeemer).markets(protocol, underlying, maturity).maturityRate);
```
robrobbins: as pointed out in the comment above...

build/creator/ZcToken.sol
```diff
@@ -61,75 +67,89 @@ contract ZcToken is Erc20, IERC5095 {
...
         if (holder == msg.sender) {
-            redeemer.authRedeem(protocol, underlying, maturity, msg.sender, receiver, previewAmount);
-            return previewAmount;
-        }
-        else {
+            IRedeemer(redeemer).authRedeem(protocol, underlying, maturity, msg.sender, receiver, previewAmount);
+        } else {
             uint256 allowed = allowance[holder][msg.sender];
-            if (allowed >= previewAmount) {
+            if (allowed < previewAmount) {
```
robrobbins, line 120: half of the actual bug fix

build/creator/ZcToken.sol
```diff
@@ -61,75 +67,89 @@ contract ZcToken is Erc20, IERC5095 {
...
         // some 5095 tokens may have custody of underlying and can can just burn PTs and transfer underlying out, while others rely on external custody
         if (holder == msg.sender) {
-            return redeemer.authRedeem(protocol, underlying, maturity, msg.sender, receiver, principalAmount);
+            return IRedeemer(redeemer).authRedeem(protocol, underlying, maturity, msg.sender, receiver, principalAmount);
         }
         else {
             uint256 allowed = allowance[holder][msg.sender];
-            if (allowed >= principalAmount) { revert Approvals(allowed, principalAmount); }
+
+            if (allowed < principalAmount) {
```
robrobbins, line 147: the other half of the bug fix

build/creator/ZcToken.sol
sourabhmarathe: I believe this also generally inline with our pattern of storing addresses and not interfaces anyways, right?

build/creator/ZcToken.sol
```diff
@@ -31,7 +31,8 @@ contract ZcToken is Erc20, IERC5095 {
         underlying = _underlying;
         maturity = _maturity;
         cToken = _cToken;
-        redeemer = IRedeemer(_redeemer);
+        // TODO historically we just store the address and hydrate via interface. should we break that pattern here?
+        redeemer = _redeemer;
```
sourabhmarathe: Unless there's a functional reason not to, I don't think we should. I did notice for some getters in solidity, it does become necessary to store the attribute as its type as hydrating an interface doesn't play well with the ABI.

build/creator/ZcToken.sol
```diff
@@ -61,75 +67,89 @@ contract ZcToken is Erc20, IERC5095 {
...
                 revert Approvals(allowed, previewAmount);
             }
             allowance[holder][msg.sender] -= previewAmount;
-            redeemer.authRedeem(protocol, underlying, maturity, holder, receiver, previewAmount); 
-            return previewAmount;
+            IRedeemer(redeemer).authRedeem(protocol, underlying, maturity, holder, receiver, previewAmount); 
         }
+
+        // TODO why use named return statements if not using them? Discuss.
+        return previewAmount;
```
sourabhmarathe: This is confusing, especially considering we generally don't use pre-named returns. I think we should update the function signature to not have identifier on the return and leave this as is.

sourabhmarathe (changes requested):
Small questions/fixes, mostly about the TODOs. Looks good

build/creator/ZcToken.sol
robrobbins: yeah

build/creator/ZcToken.sol
robrobbins: i've resisted addressing all the style conflicts introduced by this contract at once. thing is a mess.

it will likely be rounded into form as the audit fixes go in.

that being said, i'll likely remove all named returns as it's not a pattern we use anywhere, and its a quick change.

build/marketplace/MarketPlace.sol
```diff
@@ -106,13 +107,13 @@ contract MarketPlace {
     }
 
     // set the base maturity cToken exchange rate at maturity to the current cToken exchange rate
-    uint256 exchangeRate = Compounding.exchangeRate(p, market.cTokenAddr);
-    markets[p][u][m].maturityRate = exchangeRate;
+    uint256 xRate = Compounding.exchangeRate(p, market.cTokenAddr);
```
JTraversa, line 110: Better than what I had thrown into the audit which was just "rate" IIRC 👍

JTraversa (approved):
LGTM 🚢

PR #410, opened 2022-08-01
bug: check for address(0) from ecrecover in sig lib. also cleanup lib.

robrobbins:


test/swivel/Sig.sol
```diff
@@ -12,7 +12,8 @@ library Sig {
 
   error S();
   error V();
-  error SIG_LENGTH();
```
robrobbins, line 15: no other errors are `CONST_STYLE()` so remoing in favor of `Length`. The library same is `Sig` so i think context lets us know its the signature length

build/swivel/Sig.sol
```diff
@@ -24,22 +25,17 @@ library Sig {
...
-    return ecrecover(h, c.v, c.r, c.s);
-  }
+    address recovered = ecrecover(h, c.v, c.r, c.s);
 
-  /// @param h Hashed data which was originally signed
-  /// @param sig Valid ECDSA signature
-  /// @dev splitAndRecover should only be used if it is known that the resulting 
-  /// verifying bit (V) will be 27 || 28. Otherwise use recover, possibly calling split first.
-  /// @return The recovered address
-  function splitAndRecover(bytes32 h, bytes memory sig) internal pure returns (address) {
```
robrobbins, line 39: this whole method is deleted. it was a bit of legacy that got written and never used. nothing calls this.. the contracts just use `recover` with `split` being used by the Fakes and Unit tests.

robrobbins:
image

managed to force `ecrecover to return `address(0)` and write a spec that shows we now catch it (...failsOnAddress0)

pkg/testing/sig_test.go
```diff
@@ -69,6 +69,47 @@ func (s *sigTestSuite) TestRecover() {
...
+	if vrs.V < 27 {
+		vrs.V += 27
+	}
+
+	// mess with the r such that ecrecover will fail
+	vrs.R = [32]byte{}
+
+	_, err = s.Dep.SigFake.RecoverTest(nil, hash, vrs)
+
+	assert.NotNil(err)
```
robrobbins, line 110: if we comment out the new line that reverts on `address(0)` this spec will fail.

i.e -- it just returns `0x0...` and `err` is nil here

test/fakes/Sig.sol
```diff
@@ -10,35 +10,42 @@ library Sig {
...
-  /// @param sig Valid ECDSA signature
-  /// @dev splitAndRecover should only be used if it is known that the resulting 
-  /// verifying bit (V) will be 27 || 28. Otherwise use recover, possibly calling split first.
-  /// @return The recovered address
-  function splitAndRecover(bytes32 h, bytes memory sig) internal pure returns (address) {
-    (uint8 v, bytes32 r, bytes32 s) = split(sig);
+    address recovered = ecrecover(h, c.v, c.r, c.s);
 
-    return ecrecover(h, v, r, s);
+    if (recovered == address(0)) {
```
robrobbins, line 34: if this check is not present the `...failOn0Address` spec fails.

sourabhmarathe (commented):
LGTM

sourabhmarathe (approved):

PR #411, opened 2022-08-02
bug: Vault Tracker set authorized caller as MarketPlace

robrobbins:


build/creator/Creator.sol
```diff
@@ -37,8 +37,12 @@ contract Creator {
     string calldata s,
     uint8 d
   ) external authorized(marketPlace) returns (address, address) {
+    if (marketPlace == address(0)) {
+      revert Exception(34, 0, 0, marketPlace, address(0));
+    }
+
```
robrobbins, line 43: this is a new addition.

so i need to add a `creatortesting` suite next as there's not one yet. i'm going to follow this PR with that one however as it would bloat this one to _real damn big_

JTraversa (approved):
Solid.

Like a 🤘 "Rock"

PR #419, opened 2022-08-05
feat: implement changes for fee structure

robrobbins:


test/swivel/Swivel.sol
```diff
@@ -23,6 +23,8 @@ contract Swivel {
   string constant public NAME = 'Swivel Finance';
   string constant public VERSION = '3.0.0';
   uint256 constant public HOLD = 3 days;
+  /// @dev point in time at which a fee change may take place
+  uint256 public feeChange;
```
sourabhmarathe, line 27: I like the `@dev` doc here, might be a good idea to expand to other non-obvious contract attributes as well, but that's a future thing.

test/swivel/Swivel.sol
```diff
@@ -499,17 +524,35 @@ contract Swivel {
...
 
-    feenominators[i] = d;
+    if (block.timestamp < feeChange) {
+      revert Exception(36, block.timestamp, feeChange, address(0), address(0));
+    }
+
+    for (uint256 i; i < 4;) {
+      if (f[i] < MIN_FEENOMINATOR) {
+        revert Exception(18, f[i], MIN_FEENOMINATOR, address(0), address(0));
+      }
```
sourabhmarathe, line 542: This seems like `admin` input sanitation. That said, I can definitely see the merits of including it to avoid a fat-finger breaking the protocol for a bit.

sourabhmarathe (approved):
Approve, some questions/suggestions but nothing to stop the click of the button.

test/swivel/Swivel.sol
robrobbins: i agree that the MIN can be done away with. Didn't we remove it from Illuminate?

anyway - a product question that we can come back to - it's been there since we introduced the fees

robrobbins:
addressed: https://github.com/Swivel-Finance/gost/pull/419

PR #421, opened 2022-08-09
feat: give vault tracker a rates method which will not return a 0 mat…

robrobbins:
…urity rate


test/tokens/ZcToken.sol
```diff
@@ -42,8 +42,10 @@ contract ZcToken is Erc20, IERC5095 {
             return 0;
         }
 
-        // NOTE we weren't even using the pre-hydrated redeemer in most of these anyway...
-        return principalAmount * IRedeemer(redeemer).exchangeRate(protocol, cToken) / IRedeemer(redeemer).markets(protocol, underlying, maturity).maturityRate;
+        // note that maturity rate (mRate) == exchangeRate (xRate) if the mRate would have been 0 (see VaultTracker.rates)
+        (uint256 mRate, uint256 xRate) = IRedeemer(redeemer).rates(protocol, underlying, maturity);
+
+        return principalAmount * xRate / mRate;
```
sourabhmarathe, line 48: Nit: instead of `xRate` can we use the full word (spotRate?), same for `mRate`. Up to you if this change is worth doing, but I think it would help with legibility of the codebase.

test/vaulttracker/VaultTracker.sol
```diff
@@ -246,6 +211,14 @@ contract VaultTracker {
     return true;
   }
 
+  /// @notice Return both the current maturityRate if it's > 0 (or exchangeRate in its place) and the Compounding exchange rate
+  /// @dev While it may seem unnecessarily redundant to return the exchangeRate twice, it prevents many kludges that would otherwise be necessary to guard it
+  /// @return maturityRate, exchangeRate if maturityRate > 0, exchangeRate, exchangeRate if not.
+  function rates() public view returns (uint256, uint256) {
+    uint256 exchangeRate = Compounding.exchangeRate(protocol, cTokenAddr);
+    return maturityRate > 0 ? (maturityRate, exchangeRate) : (exchangeRate, exchangeRate);
```
sourabhmarathe: I noticed we use these operators `?` etc. instead of `if-else` statements. If they aren't more efficient, I think we should make it a practice to avoid these ternary operators as they are a bit harder to read. That said, this is a nit and something that can be addressed another time.

sourabhmarathe (dismissed):
Just a few nits/questions, nothing much to add. LGTM overall

test/vaulttracker/VaultTracker.sol
robrobbins: ternary is the correct operator for a truly binary decision, so i support their use.

effeciency would be a determining factor but i'm pretty sure it results in the same bytecode. could be worth a look.

test/tokens/ZcToken.sol
robrobbins: spotRate does not exist in the codebase and i'm not going to introduce another word. its pretty clearly marked these are maturity rate and exchange rate. 

`maturityRate` would clash with the global. i damn sure aint using `_maturityRate` so `mRate` works for me.

build/creator/VaultTracker.sol
```diff
@@ -246,6 +211,14 @@ contract VaultTracker {
     return true;
   }
 
+  /// @notice Return both the current maturityRate if it's > 0 (or exchangeRate in its place) and the Compounding exchange rate
+  /// @dev While it may seem unnecessarily redundant to return the exchangeRate twice, it prevents many kludges that would otherwise be necessary to guard it
+  /// @return maturityRate, exchangeRate if maturityRate > 0, exchangeRate, exchangeRate if not.
+  function rates() public view returns (uint256, uint256) {
+    uint256 exchangeRate = Compounding.exchangeRate(protocol, cTokenAddr);
```
JTraversa, line 218: Edit: Never miiiind

test/tokens/ZcToken.sol
JTraversa: i too hate reading mRate and xRate but dont have better solution, and agree that itd be `exchangeRate` and `maturityRate` or an alternative to `maturityRate`

test/vaulttracker/VaultTracker.sol
JTraversa: Im neutral on this overall.

1. I agree its slightly harder to read which can be important in solidity where were throwing it at auditors
2. But those same auditors have expectations based on lines of code submitted and semi charge based on that so there is incentive to reduce LOC overall lol

PR #428, opened 2022-08-22
gas opts / styles for parity with illuminate. some audit ticket gas o…

robrobbins:
…pts as well (195)


build/marketplace/MarketPlace.sol
```diff
@@ -63,12 +63,13 @@ contract MarketPlace is IMarketPlace {
...
   /// @param n Name of the new market zcToken
+  /// @dev the memory allocation of `s` is for alleviating STD err, there's no clearly superior scoping or abstracting alternative.
   /// @param s Symbol of the new market zcToken
   function createMarket(
     uint8 p,
     uint256 m,
     address c,
-    string memory n,
+    string calldata n,
     string memory s
```
robrobbins, line 73: i, ofc, wanted to declare both as `calldata` but STD won't have it here. and there is no superior way to scope out of this one (the vars are used throughout) and making a new method to do half this functionality would erase the gas savings 
so... this is what we have

the only other option is to get rid of "symbol" and just use "name" twice but that's not _exactly_ standard practice so
:shrug:

sourabhmarathe (approved):
🚢

build/marketplace/MarketPlace.sol
JTraversa: could have them both as bytes and use a bytes concat library to give it a `- maturity timestamp` but i think thats probably overkill hah

PR #438, opened 2022-09-01
cancel does not need sig components and hashing overhead

robrobbins:
this is a suggestion from: https://github.com/code-423n4/2022-07-swivel-findings/issues/165

tl;dr - when cancelling an order(s) you don't need to signature check - just check that msg.sender is order.maker

if we are comforable with that -- it's a bit less code and processing. no strong feelings from me either way.


build/swivel/Interfaces.sol
```diff
@@ -9,7 +9,7 @@ import './Sig.sol';
 interface ISwivel {
   function initiate(Hash.Order[] calldata, uint256[] calldata, Sig.Components[] calldata) external returns (bool);
   function exit(Hash.Order[] calldata, uint256[] calldata, Sig.Components[] calldata) external returns (bool);
-  function cancel(Hash.Order calldata, Sig.Components calldata) external returns (bool);
+  function cancel(Hash.Order[] calldata) external returns (bool);
```
robrobbins, line 12: the V3 branch did not have the update that let `cancel()` be called with an array of orders

build/swivel/Swivel.sol
```diff
@@ -440,18 +440,22 @@ contract Swivel is ISwivel {
   }
 
   /// @notice Allows a user to cancel an order, preventing it from being filled in the future
-  /// @param o Order being cancelled
-  /// @param c Components of a valid ECDSA signature
-  function cancel(Hash.Order calldata o, Sig.Components calldata c) external returns (bool) {
-    bytes32 hash = validOrderHash(o, c);
+  /// @param o Array of offline orders being cancelled
+  function cancel(Hash.Order[] calldata o) external returns (bool) {
```
robrobbins, line 444: as stated above, take an array
note we can remove the components if not checking the sig

robrobbins:
also gatta get a spec on this. will do before merging.

JTraversa (approved):
Good catch by the warden, was easy to just assume is necessary

🚢

Issue #439, opened 2022-09-06
Lido


JTraversa:
We had commented out Lido for some later considerations awhile back, and never re-introduced it.

Its 100% worth taking the bit of time to add Lido and ensure we're good to go for the next year or so.

That said, we need to address the three interfaces for integrations:
1. deposit -- [`wrap`](https://docs.lido.fi/contracts/wsteth/#wrap)
2. withdraw -- [`unwrap`](https://docs.lido.fi/contracts/wsteth/#unwrap)
3. exchangeRate -- [`getStETHByWstETH`](https://docs.lido.fi/contracts/wsteth/#getstethbywsteth)

Alex generally likes having `supplyRatePerBlock` to calculate APY as well, but I'm fairly sure that supplyRate will be extremely difficult to extrapolate. Its easy enough now, but varies based on tips/MEV post merge...

Other notes:

The underlying is not ETH or WETH, but stETH for wstETH.

So we can assume users have stETH ahead of time, or can write a wrapper that converts ETH->stETH through curve?

robrobbins:
1. i'll start on the integrations
2. worry about supplyRate in a sep thing

JTraversa:
ah and for the `underlying` in one of those compounding libs, it can be hardcoded to `0xae7ab96520de3a18e5e111b5eaab095312d7fe84` (stETH)

robrobbins:
oh, ic. ok. lemme look at the best place for that.

was just reading the whole:

```
_stETHAmount must be non-zero
msg.sender must approve at least _stETHAmount stETH to this contract.
msg.sender must have at least _stETHAmount of stETH.```

for wrap...

PR #440, opened 2022-09-09
Feat/lido

robrobbins:
lfirst pass thru getting Lido in as per #439 

open question: are we happy with the `const` in the `Compounding` lib for getting tthe equivalent for a Lido protocol "underlying" ?

i can always dig for some dynamic way to get that address, but, that has other questions to go with it.

open question: is the assumption about how to perform `deposit` in `Swivel.sol` correct? obvs we'll have to test that but, worth asking...

I'll push up spec updates in a follow on to this -- just wanted to get the code up for a look-see


test/libraries/Compounding.sol
```diff
@@ -43,7 +43,15 @@ interface IEulerToken {
...
 
+interface ILidoToken {
+  /// @notice Returns amount of stETH for a given amount of wstETH
+  function getStETHByWstETH(uint256) external view returns (uint256);
+  /// @dev NOTE there is no underlying asset here as Compounding stores the Lido stETH address directly
+}
+
 library Compounding {
+  // TODO can we dynamically fetch this? Other options?
+  address constant LIDO_STETH = '0xae7ab96520de3a18e5e111b5eaab095312d7fe84';
```
sourabhmarathe: This seems to be something like `aaveAddr` in Swivel.sol. It's a fixed address so it won't change.

Also we can get the stETH address with the [`stETH` method](https://etherscan.io/token/0x7f39c581f595b53c5cb19bd0b3f8da6c935e2ca0#readContract#F9) on wstETH.

test/swivel/Interfaces.sol
```diff
@@ -60,6 +60,12 @@ interface IEuler {
   function withdraw(uint256, uint256) external; // void
 }
 
+interface ILido {
+  function wrap(uint256) external returns (uint256);
+  function unwrap(uint256) external returns (uint256);
+  function getStETHByWstETH(uint256) external returns (uint256);
+}
+
```
sourabhmarathe, line 68: Re: getting stETH address, it can be retrieved using `stETH`.

See the method here:
https://etherscan.io/token/0x7f39c581f595b53c5cb19bd0b3f8da6c935e2ca0#readContract#F9

sourabhmarathe (changes requested):
One change to answer the question from the `TODO`.

test/swivel/Swivel.sol
```diff
@@ -821,6 +823,9 @@ contract Swivel is ISwivel {
       // Euler withdraw is void
       IEuler(c).withdraw(0, a);
       return true;
+    } else if (p == uint8(Protocols.Lido)) {
+      ILido wstEth = ILido(c);
+      return wstEth.unwrap(a / wstEth.getStETHByWstETH(1e18)) >= 0;
```
sourabhmarathe: I'm not sure if this is the right call - we may be able to get away with just passing the entirety of the balance or `a` (assuming `a` represents the user's amount of tokens they would like to withdraw.
```suggestion
      return wstEth.unwrap(IErc20(address(c).balanceOf(msg.sender or a)) >= 0;
```

sourabhmarathe: I could be wrong but it looks like [`unwrap`](https://docs.lido.fi/contracts/wsteth#unwrap) takes in the amount of tokens weighted staked ETH amount.

robrobbins: @JTraversa bring some prod knowledge in here... 
this pass i had assumed it was like the current yearn solution.

test/swivel/Interfaces.sol
robrobbins: yeah, as long as that is the equivalent of `underlying` then 100% going with this

JTraversa: Yea agree. I missed this on my look through and its way more consistent w/ our pattern

test/swivel/Swivel.sol
JTraversa: Unwrap seems to accept the wstETH balance as its param, in other words `shares` of the wstETH vault (compounding token amounts), rather than raw `stETH` amounts (underlying token amounts) 

So using the underlying amount wouldnt work (which is why i think is being suggested?)
_not 100% what you mean by "weighted staked eth amount"_

test/swivel/Interfaces.sol
robrobbins: ooohh. it's a public variable. that's why we didn't see it in the method list

test/swivel/Swivel.sol
sourabhmarathe: I don't understand why we don't just call `unwrap` on the amount of tokens (`a`). If `a` is the number of wstETH tokens, then this would be simpler than calculating the rate.

JTraversa: @sourabhmarathe 
Your first comment:
> (assuming a represents the user's amount of tokens they would like to withdraw)

New comment: 

> If a is the number of wstETH tokens

So thats where your confusion is.

`a` is the user's amount of underlying tokens they would like to withdraw, not the amount of `shares` (wstETH)

sourabhmarathe (approved):
LGTM

